PR #3, opened 2025-04-22
Set up basic collection infrastructure

felixfontein:
##### SUMMARY
I've used antsibull-nox for CI, since a) it's the easiest to set up, and b) I wrote it ;-) If you don't like it you can replace it later on. (I'd add integration test in a separate workflow, and only maybe later think of integrating them in antsibull-nox as well. But that's for another day anyway...)

Right now CI will fail since the files in the collection refer to community.general all over the place. I'll create another PR for that to fix it and rebase this one once that is merged.

##### ISSUE TYPE
- Test Pull Request
- Docs Pull Request

##### COMPONENT NAME
infrastructure


felixfontein:
I also noticed some files missing from community.general; I've added them in this PR.

felixfontein:
Hmmm I noticed I cannot request a review from @WaitingAtTheCrossroads yet, so here's a manual ping :)

I think with this PR you have a basic collection that you can work on. I'd probably proceed as follows:
1. Get this PR reviewed, adjusted, and merged.
2. Handle the deprecation in plugins/modules/proxmox.py. I'd basically act like this is community.general 11.0.0, i.e. apply the announced changes.
3. Release a first 0.1.0 version once everything looks good enough for a first try.
4. Polishing, experimenting with integration tests, ...
5. Get a 0.2.0 out soon, or even directly skip to 1.0.0 if it looks good enough for a first version. Then we can officially add it to Ansible, and deprecate the proxmox content in community.general.

What do you all think?

felixfontein:
@gundalow I think you can re-add the branch protections. Thanks a lot! Also it would be great if you could give more rights to @WaitingAtTheCrossroads :)

gundalow:
@WaitingAtTheCrossroads please check your email, you've got an invite for this repo

felixfontein:
@Thulium-Drake @WaitingAtTheCrossroads feel free to look at the changes and comment (or add suggestions), or if you're happy with it approve the PR.

galaxy.yml
```diff
@@ -0,0 +1,28 @@
...
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+# See https://docs.ansible.com/ansible/latest/dev_guide/collections_galaxy_meta.html
+
+namespace: community
+name: proxmox
+version: 0.0.1
+readme: README.md
+authors:
+  - YOUR NAME (github.com/YOURGITHUB)
```
WaitingAtTheCrossroads: This stands out.  It looks like many other collections, including `community.general` that we came from, use `Ansible (github.com/ansible)`.  Is that still appropriate here?

WaitingAtTheCrossroads (commented):
This looks like an **excellent** start!  😍 

I just had one comment where author metadata still has a placeholder.

Are manual runs of the integration tests also working?  I'm actually not sure how to run them properly. 😳 I can see something like `ansible-test integration --allow-unsupported` should run them, but I don't see where to point it to a test instance of proxmox.  I thought setting `PROXMOX_(HOST|PORT|USER|PASSWORD)` environment variables might work, but I still get an error that `api_host` is not set.

galaxy.yml
felixfontein: I guess that should be fine. We can also update it anytime we want.

Now thinking about it, how about adding "Ansible community" there? After all this is community stuff, not something from RH :)
```suggestion
  - Ansible community (https://github.com/ansible-community)
```

WaitingAtTheCrossroads (approved):
LGTM!

WaitingAtTheCrossroads:
Left open in case @Thulium-Drake was wanting to formally approve.  I think it will be a couple days before I will be looking at this much again anyway.

Thulium-Drake (approved):
:rocket:

felixfontein:
@WaitingAtTheCrossroads @Thulium-Drake @gundalow thanks!

PR #12, opened 2025-05-06
Add new module proxmox_backup_schedule to ensure vms backup jobs are scheduled

mmayabi:
##### SUMMARY
feature: Add new module proxmox_backup_schedule to ensure vms backup jobs are scheduled correctly. 

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME
proxmox_backup_schedule

##### ADDITIONAL INFORMATION

The proxmox_backup_schedule module modify backup jobs such as set or delete them for VMs.

### DIFF
--- a/plugins/modules/proxmox_backup_schedule.py
+++ b/plugins/modules/proxmox_backup_schedule.py
@@ -20,7 +20,7 @@ description: The module modifies backup jobs such as set or delete C(vmid).
 author:
   - "Marzieh Raoufnezhad (@raoufnezhad) <raoufnezhad@gmail.com>"
   - "Maryam Mayabi (@mmayabi) <mayabi.ahm@gmail.com>"
-
+version_added: 1.0.0
 options:
   vm_name:
     description:
--- a/plugins/modules/proxmox_backup_schedule.py
+++ b/plugins/modules/proxmox_backup_schedule.py
@@ -1,0 +1,246 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2025 Marzieh Raoufnezhad <raoufnezhad@gmail.com>
+# Copyright (c) 2025 Maryam Mayabi <mayabi.ahm at gmail.com>
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = """
+---
+module: proxmox_backup_schedule
+
+short_description: Schedule VM backups and removing them
+
+description: The module modifies backup jobs such as set or delete C(vmid).
+
+author:
+  - "Marzieh Raoufnezhad (@raoufnezhad) <raoufnezhad@gmail.com>"
+  - "Maryam Mayabi (@mmayabi) <mayabi.ahm@gmail.com>"
+
+options:
+  vm_name:
+    description:
+      - The name of the Proxmox VM.
+      - Mutually exclusive with O(vm_id).
+    type: str
+  vm_id:
+    description:
+      - The ID of the Proxmox VM.
+      - Mutually exclusive with O(vm_name).
+    type: str
+  backup_id:
+    description: The backup job ID.
+    type: str
+  state:
+    description:
+        - If V(present), ensure that VM is present in the backup job on defined O(backup_id).
+        - If V(absent), ensure that VM is not present in the backup job if O(backup_id) is defined and otherwise, it should not exist in any of the backup jobs.
+    required: true
+    choices: ["present", "absent"]
+    type: str
+
+extends_documentation_fragment:
+  - community.proxmox.proxmox.documentation
+  - community.proxmox.attributes
+  - community.proxmox.attributes.info_module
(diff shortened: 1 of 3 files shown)


felixfontein:
This PR got moved from here: https://github.com/ansible-collections/community.general/pull/9592

l00ptr:
Hi @mmayabi,

is it possible to squash or remove the tests and debug commits ? Do we need all the intermediate commits ? IMHO this is not needed since you introduce a new feature and only add new files for that feature. So could you squash the commits and adapt the commit message to describe how and why you add that new module ?

Best regards,
L.

plugins/modules/proxmox_backup_schedule.py
```diff
@@ -0,0 +1,255 @@
...
+class ProxmoxSetVMBackupAnsible(ProxmoxAnsible):
+    # Getting backup sections
+    def get_cluster_bklist(self):
+        try:
+            backupSections = self.proxmox_api.cluster.backup.get()
+        except Exception as e:
+            self.module.fail_json(msg="Getting backup sections failed: %s" % e)
+        return backupSections
+
+    def get_cluster_specific_bkjobid(self, backup_id):
```
l00ptr: what about renaming this to `get_backup(self, backup_id)` ?

plugins/modules/proxmox_backup_schedule.py
```diff
@@ -0,0 +1,255 @@
...
+"""
+
+from ansible.module_utils.basic import AnsibleModule, missing_required_lib
+from ansible_collections.community.proxmox.plugins.module_utils.proxmox import (
+    proxmox_auth_argument_spec, ProxmoxAnsible, HAS_PROXMOXER, PROXMOXER_IMP_ERR)
+
+
+class ProxmoxSetVMBackupAnsible(ProxmoxAnsible):
+    # Getting backup sections
+    def get_cluster_bklist(self):
```
l00ptr: what about renaming this to  `list_backups(self)` or may be `list_configured_backups(self)`  ?

plugins/modules/proxmox_backup_schedule.py
```diff
@@ -0,0 +1,255 @@
...
+        return backupSections
+
+    def get_cluster_specific_bkjobid(self, backup_id):
+        try:
+            specificBackupID = self.proxmox_api.cluster.backup.get(backup_id)
+        except Exception as e:
+            self.module.fail_json(msg="Getting specific backup ID failed: %s" % e)
+        return specificBackupID
+
+    def set_vmid_backup(self, backup_id, bk_id_vmids):
```
l00ptr: do we configure the backup for one VM or for a list of VMs ?

plugins/modules/proxmox_backup_schedule.py
```diff
@@ -0,0 +1,255 @@
...
+            bk_id_info = bk_id_info[0]  # Access the first item in the list
+        vms_id = bk_id_info['vmid'].split(',')
+        if str(vm_id) not in vms_id:
+            bk_id_vmids = bk_id_info['vmid'] + ',' + str(vm_id)
+            self.set_vmid_backup(backup_id, bk_id_vmids)
+            return True
+        else:
+            return False
+
+    # delete vmid from backup job
```
l00ptr: may be we can use a docstring instead of this comment ?

plugins/modules/proxmox_backup_schedule.py
```diff
@@ -0,0 +1,255 @@
...
+    def backup_absent(self, vm_id, backup_id):
+        if backup_id:
```
l00ptr: what do we check here ? if `backup_id` is `True`, if it's not `None` or if it's not an empty list ?

plugins/modules/proxmox_backup_schedule.py
```diff
@@ -0,0 +1,255 @@
...
+                        vmids.remove(str(vm_id))
+                        new_vmids = ','.join(map(str, vmids))
+                        self.set_vmid_backup(backupItem['id'], new_vmids)
+                        bkID_delvm.append(backupItem['id'])
+                    else:
+                        self.module.fail_json(msg="No more than one vmid is assigned to %s. You just can remove job." % backupItem['id'])
+            if bkID_delvm:
+                return True
+            else:
+                return False
```
l00ptr:
```suggestion
            return len(bkID_delvm)>0
```

plugins/modules/proxmox_backup_schedule.py
```diff
@@ -0,0 +1,255 @@
...
+        argument_spec=args,
+        mutually_exclusive=[('vm_id', 'vm_name')],
+        supports_check_mode=True
+    )
+
+    # Define (init) result value
+    result = dict(
+        changed=False,
+        message=''
+    )
```
l00ptr:
```suggestion
    result = dict(changed=False, message='')
```

or

```suggestion
    result = {"changed": False, "message": ''}
```

l00ptr (changes requested):

plugins/modules/proxmox_backup_schedule.py
mmayabi: This module configures the backup for one VM, but the Proxmox API endpoint expects the entire list of VM IDs, including both the existing VMs already in the backup job and the new VM ID to be added.

Thulium-Drake:
@l00ptr When I merge, I always squash all commits into one. This keeps the main repo history clean as well (this is also what community.general and the like do :-) )

PR #104, opened 2025-06-08
Fixed handling warnings

Thulium-Drake:
##### SUMMARY

I broke something when I implemented #96 I did test the KVM flow, but not the LXC flow, which broke with the error below:

```
TASK [Create new container with minimal options] **********************************************************************
fatal: [localhost]: FAILED! => changed=false 
  ansible_facts:
    discovered_interpreter_python: /usr/bin/python3
  msg: 'Unable to retrieve API task ID from node lab: ''exitstatus'''
```

I changed how the function returns the API task status and tested both flows for KVM and LXC this time ;-)

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

proxmox
##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/104-fix-warnings.yml
+++ b/changelogs/fragments/104-fix-warnings.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - proxmox module utils - fix handling warnings in LXC tasks (https://github.com/ansible-collections/community.proxmox/pull/104)
--- a/changelogs/fragments/104-fix-warnings.yml
+++ b/changelogs/fragments/104-fix-warnings.yml
@@ -1,3 +1,3 @@
 ---
 minor_changes:
-  - proxmox module utils - fix handling warnings in LXC tasks (https://github.com/ansible-collections/community.proxmox/pull/104)
+  - proxmox module utils - fix handling warnings in LXC tasks (https://github.com/ansible-collections/community.proxmox/pull/104).
--- a/plugins/module_utils/proxmox.py
+++ b/plugins/module_utils/proxmox.py
@@ -21,6 +21,7 @@ except ImportError:
 
 
 from ansible.module_utils.basic import env_fallback, missing_required_lib
+from ansible.module_utils.common.text.converters import to_native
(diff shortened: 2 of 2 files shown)


plugins/module_utils/proxmox.py
```diff
@@ -166,8 +166,7 @@ def get_vm(self, vmid, ignore_missing=False):
     def api_task_ok(self, node, taskid):
         try:
             status = self.proxmox_api.nodes(node).tasks(taskid).status.get()
-            valid_exitstatus = status['exitstatus'] == 'OK' or status['exitstatus'].startswith('WARN')
-            return status['status'] == 'stopped' and valid_exitstatus
+            return status['status'] == 'stopped' and (status['exitstatus'] == 'OK' or status['exitstatus'].startswith('WARN'))
```
felixfontein: How about:
```suggestion
            exitstatus = to_native(status.get('exitstatus') or '')
            return status['status'] == 'stopped' and (exitstatus == 'OK' or exitstatus.startswith('WARN'))
```
That way even if `exitstatus` doesn't exist, you'll end up with a string (the empty one).

changelogs/fragments/104-fix-warnings.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+  - proxmox module utils - fix handling warnings in LXC tasks (https://github.com/ansible-collections/community.proxmox/pull/104)
```
felixfontein:
```suggestion
  - proxmox module utils - fix handling warnings in LXC tasks (https://github.com/ansible-collections/community.proxmox/pull/104).
```

plugins/module_utils/proxmox.py
Thulium-Drake: Right, that makes sense and is even more future proof in case the PVE guys decide to change something on that end, thanks for the tip! :-)

Thulium-Drake:
I have tested the proxmox part better this time around, and as it's currently broken, I'll merge it by the end of tomorrow, so we can release 1.0.1 before 1.0.0 hits the streets proper :sweat_smile:

PR #105, opened 2025-06-08
Allow force deletion of LXC containers when they are running

Thulium-Drake:
##### SUMMARY

When testing the fix for #104 I had to delete a lot of containers and noticed that the module does not actually allows to force remove containers. The PVE CLI and API do allow this by adding the `force`/`-f` parameter to the request/command.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

proxmox.py
##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/proxmox.py
+++ b/plugins/modules/proxmox.py
@@ -227,8 +227,9 @@ options:
   force:
     description:
       - Forcing operations.
-      - Can be used only with states V(present), V(stopped), V(restarted).
+      - Can be used only with states V(absent), V(present), V(stopped), V(restarted).
       - With O(state=present) force option allow to overwrite existing container.
+      - With O(state=absent) force option will force stop the container before deletion.
       - With states V(stopped), V(restarted) allow to force stop instance.
       - When specifying O(force), O(update) must not be specified.
     type: bool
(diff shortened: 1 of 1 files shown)


changelogs/fragments/105-allow-lxc-force-delete.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+  - proxmox - allow force deletion of LXC containers (https://github.com/ansible-collections/community.proxmox/pull/105)
```
felixfontein:
```suggestion
  - proxmox - allow force deletion of LXC containers (https://github.com/ansible-collections/community.proxmox/pull/105).
```

plugins/modules/proxmox.py
```diff
@@ -227,8 +227,9 @@
   force:
     description:
       - Forcing operations.
-      - Can be used only with states V(present), V(stopped), V(restarted).
+      - Can be used only with states V(absent), V(present), V(stopped), V(restarted).
```
felixfontein:
```suggestion
      - Can be used only with states V(absent), V(present), V(stopped), and V(restarted).
```

felixfontein (commented):
Looks good as far as I can judge.

PR #108, opened 2025-06-18
Inventory plugins: remove deprecated disable_lookups parameter (which was set to its default anyway)

felixfontein:
##### SUMMARY
`Templar.template`'s parameter `disable_lookups` has been deprecated in ansible-core 2.19 (https://github.com/ansible/ansible/blob/34abc83822a4ee118cc813862ecdb8d2a6a1538b/lib/ansible/template/__init__.py#L273-L279). Inventory plugins have been using it (probably copied around from a common source), but they always set it to `False`, which happened to be the default since at least 2017 (I didn't bother to check back even more).

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
inventory plugins

### DIFF
--- a/changelogs/fragments/108--disable_lookups.yml
+++ b/changelogs/fragments/108--disable_lookups.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - "proxmox inventory plugin - avoid using deprecated option when templating options (https://github.com/ansible-collections/community.proxmox/pull/108)."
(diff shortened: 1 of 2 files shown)


Thulium-Drake:
Well, since I made the first version of the module after 2017 and it's based on the Foreman inventory I'm pretty sure it was copied around ;-)

Thanks for the fix!

felixfontein:
yw! I grepped around in my ansible_collections/community/ directory and saw several instances in different collections, so I fixed them all. The fix is identical anyway ;)

Issue #85, opened 2025-04-11
Adding VM to HA group using proxmox module


jaywor1:
### Summary

Hi,
I am trying to define VMs in proxmox in ansible repository but one thing is kinda missing for full configuration that is currently manually set in proxmox and that is HA group.

I would like to see community.general.proxmox.* configuration option for VM HA group.

It's solvable using direct api calls, but i think it would be nicer to have it configured in one place

### Issue Type

Feature Idea

### Component Name

proxmox

### Additional Information

```yaml (paste below)

```

### Code of Conduct

- [x] I agree to follow the Ansible Code of Conduct

russoz:
@felixfontein this should probably be moved to the proxmox collection

jaywor1:
Thanks for implementing this feature <3

PR #111, opened 2025-06-20
new module: proxmox_cluster_ha_groups

commonism:
##### SUMMARY
This modules allows creating/deleting cluster HA groups.
required for #85

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
proxmox_cluster_ha_group

##### ADDITIONAL INFORMATION
I'll throw in unit tests if the code is acceptable.

### DIFF
--- a/plugins/modules/proxmox_cluster_ha_groups.py
+++ b/plugins/modules/proxmox_cluster_ha_groups.py
@@ -1,0 +1,191 @@
+#!/usr/bin/python
+
+# Copyright: (c) 2025, Markus Kötter <koetter@cispa.de>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+from __future__ import (absolute_import, division, print_function)
+
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+---
+module: proxmox_cluster_ha_groups
+
+short_description: Management of HA groups in roxmox VE Cluster
+
+version_added: "1.0.2"
+
+description: 
+  - Configure HA groups via /cluster/ha/groups
+
+options:
+    state:
+        description: create or delete
+        required: true
+        choices: ['present', 'absent']
+        type: str
+    name:
+        description: group name
+        required: true
+        type: str
+    comment:
+        description: Description
+        required: false
+        type: str
+    nodes:
+        description: List of cluster node members, where a priority can be given to each node. A resource bound to a group will run on the available nodes with the highest priority. If there are more nodes in the highest priority class, the services will get distributed to those nodes. The priorities have a relative meaning only. The higher the number, the higher the priority.
+        required: false
+        type: bool
+    nofailback:
+        description: The CRM tries to run services on the node with the highest priority. If a node with higher priority comes online, the CRM migrates the service to that node. Enabling nofailback prevents that behavior.
+        required: false
+        type: bool
+    restricted:
+        description: Resources bound to restricted groups may only run on nodes defined by the group. The resource will be placed in the stopped state if no group node member is online. Resources on unrestricted groups may run on any cluster node if all group members are offline, but they will migrate back as soon as a group member comes online. One can implement a 'preferred node' behavior using an unrestricted group with only one member.
+        required: False
+        type: bool
+
+author:
+    - Markus Kötter (@commonism)
+'''
+
+EXAMPLES = r'''
+- name: Create HA group
+  community.proxmox.proxmox_cluster_ha_groups:
+    api_host: "{{ ansible_host }}"
+    api_password: "{{ proxmox_root_pw | default(lookup('ansible.builtin.env', 'PROXMOX_PASSWORD', default='')) }}"
+    api_user: root@pam
+
+    state: "present"
+    name: ha0
+    comment: yes
+    nodes: node0:0,node1:1
+    nofailback: true
+    restricted: false
+
+- name: Delete HA group
+  community.proxmox.proxmox_cluster_ha_groups:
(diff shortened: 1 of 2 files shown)


Thulium-Drake (approved):
Looks good to me, I sadly don't have an HA cluster on which I can test these. But I think this will work just fine.

On the subject of tests, unit tests for now should suffice for allowing CI to pass. We'll pick up integration tests once I've wrapped up building the infra for it (#7)

plugins/modules/proxmox_cluster_ha_groups.py
```diff
@@ -0,0 +1,191 @@
...
+
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+---
+module: proxmox_cluster_ha_groups
+
+short_description: Management of HA groups in roxmox VE Cluster
+
+version_added: "1.0.2"
```
felixfontein: Features can only be added to new minor versions, not to bugfix versions. So pick the next expected minor release:
```suggestion
version_added: "1.1.0"
```

plugins/modules/proxmox_cluster_ha_groups.py
```diff
@@ -0,0 +1,191 @@
...
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+from __future__ import (absolute_import, division, print_function)
+
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+---
+module: proxmox_cluster_ha_groups
+
+short_description: Management of HA groups in roxmox VE Cluster
```
felixfontein: I think a P is missing here :)
```suggestion
short_description: Management of HA groups in Proxmox VE Cluster
```

plugins/modules/proxmox_cluster_ha_groups.py
```diff
@@ -0,0 +1,191 @@
...
+
+version_added: "1.0.2"
+
+description: 
+  - Configure HA groups via /cluster/ha/groups
```
felixfontein: Generally all descriptions (except `short_description`) should have paragraphs with full sentences.
```suggestion
  - Configure HA groups via C(/cluster/ha/groups).
```
Also note that you should use Ansible markup (see https://docs.ansible.com/ansible/devel/dev_guide/developing_modules_documenting.html#linking-within-module-documentation and the sections below it for details).

plugins/modules/proxmox_cluster_ha_groups.py
```diff
@@ -0,0 +1,191 @@
...
+
+options:
+    state:
+        description: create or delete
```
felixfontein:
```suggestion
        description: Whether the HA groups should be there (created if missing) or not (deleted if they exist).
```

plugins/modules/proxmox_cluster_ha_groups.py
```diff
@@ -0,0 +1,191 @@
...
+    comment:
+        description: Description
+        required: false
+        type: str
+    nodes:
+        description: List of cluster node members, where a priority can be given to each node. A resource bound to a group will run on the available nodes with the highest priority. If there are more nodes in the highest priority class, the services will get distributed to those nodes. The priorities have a relative meaning only. The higher the number, the higher the priority.
+        required: false
+        type: bool
+    nofailback:
+        description: The CRM tries to run services on the node with the highest priority. If a node with higher priority comes online, the CRM migrates the service to that node. Enabling nofailback prevents that behavior.
```
felixfontein:
```suggestion
        description: The CRM tries to run services on the node with the highest priority. If a node with higher priority comes online, the CRM migrates the service to that node. Setting O(nofailback=true) prevents that behavior.
```

plugins/modules/proxmox_cluster_ha_groups.py
```diff
@@ -0,0 +1,191 @@
...
+RETURN = r'''
+# These are examples of possible return values, and in general should use other names for return values.
+old_groups:
+    description: The original name param that was passed in.
+    type: list
+    returned: always
+new_groups:
+    description: The output message that the test module generates.
+    type: list
+    returned: when changed
```
felixfontein: The `RETURN` section needs to be filled out with real data. If there is no return value, use `RETURN = r'''#'''` or something like that.

plugins/modules/proxmox_cluster_ha_groups.py
```diff
@@ -0,0 +1,191 @@
...
+    name = module.params['name']
+    comment = module.params['comment']
+    nodes = module.params['nodes']
+    nofailback = module.params['nofailback']
+    restricted = module.params['restricted']
+    try:
+        groups = proxmox._get()
+
+        if module.params["state"] == "present":
+            r = proxmox.create(groups, name, comment, nodes, nofailback, restricted)
```
felixfontein: I would use a better name than `r`, for example `changed`.

felixfontein (commented):
Thanks for your contribution! Some comments, mostly about the documentation:

plugins/modules/proxmox_cluster_ha_groups.py
```diff
@@ -0,0 +1,191 @@
...
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+---
+module: proxmox_cluster_ha_groups
+
+short_description: Management of HA groups in roxmox VE Cluster
+
+version_added: "1.0.2"
+
```
felixfontein, line 18: You also need to include the correct doc fragments, and set the `check_mode` and `diff` attributes. See the other modules for examples.

Thulium-Drake:
:tada:

Issue #113, opened 2025-06-23
[community.proxmox.proxmox] 500 Internal Server Error: SSH public key validation error


RRohi:
##### SUMMARY

After fiddling with the community.proxmox.promox module for hours to get it working, it looks like I've made it to the final boss. For whatever reason, I'm getting the following error: ``An error occurred: 500 Internal Server Error: SSH public key validation error``.
Not sure if this is a bug or a bug originating from this module but the module triggers it.

- I have installed my internal CA root certificates in the Ansible host and Proxmox.
- I changed the proxmox pve certificate. GUI doesn't warn about insecure certificate anymore, but I don't know if the API has a different certificate. Probably not.
- SSH keypair is generated using ssh-keygen.
- At first I didn't have pip-system-certs pip package installed, but I installed it and added my CA cert to ``~/PMX/lib/python3.11/site-packages/certifi/cacert.pem``. Didn't help.

I don't know what is being checked against what.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

community.proxmox.proxmox

##### ANSIBLE VERSION

```paste below
ansible [core 2.14.18]
  config file = /root/Projects/Ansible/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.11.2 (main, Apr 28 2025, 14:11:48) [GCC 12.2.0] (/usr/bin/python3)
  jinja version = 3.1.2
  libyaml = True
```

##### COLLECTION VERSION

```paste below
# /root/.ansible/collections/ansible_collections
Collection        Version
----------------- -------
community.proxmox 1.0.1
```

##### CONFIGURATION

```paste below
ANSIBLE_HOME(/root/Projects/Ansible/ansible.cfg) = /root/.ansible
COLLECTIONS_ON_ANSIBLE_VERSION_MISMATCH(/root/Projects/Ansible/ansible.cfg) = warning
CONFIG_FILE() = /root/Projects/Ansible/ansible.cfg
DEFAULT_BECOME(/root/Projects/Ansible/ansible.cfg) = True
DEFAULT_FORKS(/root/Projects/Ansible/ansible.cfg) = 5
[...]

PR #117, opened 2025-06-24
Added missing examples of some parameters.

RRohi:
##### SUMMARY

Troubleshooted my playbook for way too long through trial and error because there weren't any examples of pubkey, api_token_id, api_token_secret parameters.
Decided to add examples for next people who might struggle with these.

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

community.proxmox.proxmox

##### ADDITIONAL INFORMATION

N/A

```paste below
N/A
```

### DIFF
--- a/plugins/modules/proxmox.py
+++ b/plugins/modules/proxmox.py
@@ -590,7 +590,7 @@ EXAMPLES = r"""
     api_host: node1
     state: absent
 
-- name: Create a new container automatically selecting the next available vmid using a non-root token.
+- name: Create a new container automatically selecting the next available vmid using a non-root API token.
   community.proxmox.proxmox:
     node: 'uk-mc02'
     api_token_id: 'svc-tkn'
--- a/plugins/modules/proxmox.py
+++ b/plugins/modules/proxmox.py
@@ -590,7 +590,9 @@ EXAMPLES = r"""
     api_host: node1
     state: absent
 
-- name: Create a new container automatically selecting the next available vmid using a non-root API token.
+- name: >-
+    Create a new container automatically selecting the next available vmid
+    using a non-root API token
   community.proxmox.proxmox:
     node: 'uk-mc02'
     api_token_id: 'svc-tkn'
--- a/plugins/modules/proxmox.py
+++ b/plugins/modules/proxmox.py
@@ -615,7 +615,7 @@ EXAMPLES = r"""
     password: '123456'
     hostname: 'example.org'
     ostemplate: 'local:vztmpl/ubuntu-14.04-x86_64.tar.gz'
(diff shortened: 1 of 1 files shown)


RRohi:
My struggle is documented here.
https://github.com/ansible-collections/community.proxmox/issues/113

By the time I created the issue, I had already spent several hours getting token authentication to work through trial and error.

plugins/modules/proxmox.py
```diff
@@ -589,6 +589,28 @@
     api_password: 1q2w3e
     api_host: node1
     state: absent
+
+- name: Create a new container automatically selecting the next available vmid using a non-root API token.
```
felixfontein: This line is a bit long, which on the HTML docs would force readers on narrow-width screens to scroll to the right. It's better to break this up into multiple lines. (Also there should not be a period at the end.)
```suggestion
- name: >-
    Create a new container automatically selecting the next available vmid
    using a non-root API token
```

plugins/modules/proxmox.py
```diff
@@ -589,6 +589,28 @@
...
+    node: 'uk-mc02'
+    api_token_id: 'svc-tkn'
+    api_token_secret: '81c09a2a-0359-4ba1-8153-8cb3cd02509b'
+    api_user: 'remoteapiuser@pam'
+    api_host: 'node1'
+    password: '123456'
+    hostname: 'example.org'
+    ostemplate: 'local:vztmpl/ubuntu-14.04-x86_64.tar.gz'
+
+- name: Create a new container automatically selecting the next available vmid and providing a public key for the root account the Ansible host can use to connect to the new container.
```
felixfontein:
```suggestion
- name: >-
    Create a new container automatically selecting the next available vmid
    and providing a public key for the root account the Ansible host can use
    to connect to the new container
```

plugins/modules/proxmox.py
```diff
@@ -589,6 +589,28 @@
...
+  community.proxmox.proxmox:
+    node: 'uk-mc02'
+    api_user: 'root@pam'
+    api_password: '1q2w3e'
+    api_host: 'node1'
+    password: '123456'
+    hostname: 'example.org'
+    ostemplate: 'local:vztmpl/ubuntu-14.04-x86_64.tar.gz'
+    pubkey: 'ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIIU3tbNDdY6OWuYgJpjXSm3LFD6Y1lrIhZVQVz/FhBWA ansibleuser@ansiblehost'
```
felixfontein: To avoid having a "real" public key in here (where we don't know the private key of), and to shorten this a bit, let's simply replace parts with `...`:
```suggestion
    pubkey: 'ssh-ed25519 AAAAC3NzaC1...hBWA ansibleuser@ansiblehost'
```

Issue #118, opened 2025-07-01
proxmox_cluster Provide a sane error message if cluster name is too long


gyptazy:
##### SUMMARY
The module provides the option to provide a cluster name. Unfortunately, this can only be 15 chars. When providing longer cluster names, the underlying API of Proxmox just returns an error code 400 without a  reasonable error message which may bring up confusions. 

##### ISSUE TYPE
- Feature Idea

Validate the given length of the cluster name and provide a reasonable error message.

##### COMPONENT NAME
proxmox_cluster

PR #119, opened 2025-07-01
feature: Add validation of the cluster name lenght (which is limited).

gyptazy:
##### SUMMARY
Proxmox cluster names are limited in the lengths and may only provide 15 chars. Longer names result on an obscure error error, where the API simply provides an error 400 without any further information. Therefore, we should pre-validate the length of the given cluster name and fail if it's too long.

Fixes: #118

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
proxmox_cluster


Thulium-Drake:
Thanks for the fix! I didn't know that was a thing ;-)

Thulium-Drake (approved):

gyptazy:
> Thanks for the fix! I didn't know that was a thing ;-)

Me neither, until a mate encountered this today 🤣 

Thanks for merging that quick @Thulium-Drake!

PR #125, opened 2025-07-06
Make cluster creation idempotent

Thulium-Drake:
##### SUMMARY

This makes the creation of a cluster idempotent, up until now the cluster module just failed when a cluster was already present. It will now check if the cluster is the same as the cluster that is requested. If so, return ok with no change.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

proxmox_cluster

### DIFF
--- a/plugins/modules/proxmox_cluster.py
+++ b/plugins/modules/proxmox_cluster.py
@@ -99,24 +99,29 @@ class ProxmoxClusterAnsible(ProxmoxAnsible):
         cluster_name = self.module.params.get("cluster_name") or self.module.params.get("api_host")
         payload = {"clustername": cluster_name}
 
+        # Get cluster data
+        cluster_data = self.proxmox_api.cluster.config.totem.get()
+        # If we have data, check if we're already member of the desired cluster or a different one
+        if 'cluster_name' in cluster_data and cluster_data['cluster_name'] is not None:
+            if cluster_data['cluster_name'] == cluster_name:
+                self.module.exit_json(changed=False, msg="Cluster '{}' already present.".format(cluster_name), cluster=cluster_name)
+            else:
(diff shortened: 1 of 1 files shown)


Thulium-Drake:
@gyptazy I made some edits to the module, want to take a look and give them a try? :-)

gyptazy:
Hey @Thulium-Drake 

> @gyptazy I made some edits to the module, want to take a look and give them a try? :-)

Nice! Had exactly that on my list for the next weekend :) it already came up on Twitter and a customer. So, thanks a lot!

I’ll have a look at it tomorrow, thanks for letting me now!

Have a nice evening!
gyptazy

plugins/modules/proxmox_cluster.py
```diff
@@ -99,24 +100,29 @@ def cluster_create(self):
         cluster_name = self.module.params.get("cluster_name") or self.module.params.get("api_host")
         payload = {"clustername": cluster_name}
 
+        # Get cluster data
+        cluster_data = self.proxmox_api.cluster.config.totem.get()
+        # If we have data, check if we're already member of the desired cluster or a different one
+        if 'cluster_name' in cluster_data and cluster_data['cluster_name'] is not None:
```
gyptazy: This fails with:
```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TypeError: argument of type 'NoneType' is not iterable
```

If we do not have any cluster initialized, `cluster_data` is type `None` which lets this part fail. We can easily solve this by replacing the L106 with:
```
if cluster_data and 'cluster_name' in cluster_data and cluster_data['cluster_name'] is not None:
```

Here, we're checking if `cluster_data` is existing before.

Thulium-Drake: Oh, right, I must have missed that :sweat_smile:  thanks for the suggestion. I've added it :-)

Issue #128, opened 2025-07-11
proxmox_node New module for managing Nodes objects within a cluster


gyptazy:
##### SUMMARY
As a global product owner of virtualization, with the sub product Proxmox, I want to have two teams involved in the Proxmox cluster management with different responsibilities. This also involves different scopes of administration and usage. From a technical perspective, both ones use Ansible but:
* Linux Core Team: May access the hypervisors by SSH and can also manage the systems with Ansible by SSH. This is required to perform more "underlying" and core things of the systems.
* Proxmox Core Team: This team only manages the Proxmox related content. This means, they ensure Proxmox is usable, can set up and manage VMs, Containers, Storage connections etc. - so basically everything that is possible by the Proxmox API.

While the Linux team can easily utilize everything by SSH (including manipulating files etc.), the Proxmox core team must rely on the Proxmox API. Therefore, it would be great to have several options available as a `proxmox_node` module which also supports some kind of DPM & power management (also in addition to my [ProxLB](https://github.com/gyptazy/ProxLB) project), DNS settings and more important certificate and license management. While we already have `proxmox_node_info` in the collections, we still miss a module to manage things directly on the node itself. Managing the licenses for 200+ nodes in an automated way with Ansible and the Proxmox API would probably be very welcomed.

##### DEVELOPMENT
A first draft of the module `proxmox_node` will be pushed quickly.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
proxmox_node

##### ADDITIONAL INFORMATION

```yaml
- name: Start a Proxmox VE Node
  community.proxmox.node:
    api_host: proxmoxhost
    api_user: root@pam
    api_password: password123
    validate_certs: false
    node_name: de-cgn01-virt01
    power_state: present
- name: Update SSL certificates on a Proxmox VE Node
  community.proxmox.node:
    api_host: proxmoxhost
    api_user: root@pam
    api_password: password123
    validate_certs: false
    node_name: de-cgn01-virt01
    certificates:
        key: /opt/ansible/key.pem
        cert: /opt/ansible/cert.pem
        state: present
        force: false
- name: Place a subscription license on a Proxmox VE Node
  community.proxmox.node:
    api_host: proxmoxhost
    api_user: root@pam
    api_password: password123
    validate_certs: false
    node_name: de-cgn01-virt01
    subscription:
        state: present
        key: ABCD-EFGH-IJKL-MNOP-QRST-UVWX-YZ0123456789
```

PR #129, opened 2025-07-11
new module: proxmox_node for managing a Proxmox VE node

gyptazy:
feature: Add new module proxmox_node for managing a Proxmox VE node

##### SUMMARY
This module brings in the possibilities to adjust the following sections by the API:
 * Manage license subscriptions
 * Manage power states
 * Manage DNS settings
 * Manage x509 certificates

This might become important when a teams's scope is limited to the Proxmox API and cannot manage the systems by SSH.

Fixes: #128
Sponsored-by: credativ GmbH (https://credativ.de)

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
proxmox_node

##### ADDITIONAL INFORMATION

```paste below
- name: Start a Proxmox VE Node
  community.proxmox.node:
    api_host: proxmoxhost
    api_user: root@pam
    api_password: password123
    validate_certs: false
    node_name: de-cgn01-virt01
    power_state: present
- name: Update SSL certificates on a Proxmox VE Node
  community.proxmox.node:
    api_host: proxmoxhost
    api_user: root@pam
    api_password: password123
    validate_certs: false
    node_name: de-cgn01-virt01
    certificates:
        key: /opt/ansible/key.pem
        cert: /opt/ansible/cert.pem
        state: present
        force: false
- name: Place a subscription license on a Proxmox VE Node
  community.proxmox.node:
    api_host: proxmoxhost
    api_user: root@pam
    api_password: password123
    validate_certs: false
    node_name: de-cgn01-virt01
    subscription:
        state: present
        key: ABCD-EFGH-IJKL-MNOP-QRST-UVWX-YZ0123456789
```

### DIFF
--- a/plugins/modules/proxmox_node.py
+++ b/plugins/modules/proxmox_node.py
@@ -1,0 +1,442 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2025, Florian Paul Azim Hoberg (@gyptazy) <florian.hoberg@credativ.de>
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = r"""
+module: proxmox_node
+version_added: 1.2.0
+short_description: Manage Proxmox VE nodes
+description:
+  - Manage the Proxmox VE nodes itself.
+author: Florian Paul Azim Hoberg (@gyptazy)
+attributes:
+  check_mode:
+    support: full
+  diff_mode:
+    support: none
+options:
+  node_name:
+    description:
+      - The targeted node to perform actions on.
+    type: str
+    required: true
+  power_state:
+    description:
+      - Manages the power state of the node.
+    type: str
+    required: false
+    choices: ["online", "offline"]
+  certificates:
+    description:
+      - Manages the x509 certificates of the node.
+    type: dict
+    suboptions:
+      cert:
+        description:
+          - The public certificate file (including chain) in PEM format.
+        type: str
+      key:
+        description:
+          - The private key file in PEM format.
+        type: str
+      state:
+        description:
+          - Defines the actions for the certificate.
+        choices: ["present", "absent"]
+        type: str
+      restart:
+        description:
+          - Restart pveproxy to rehash the new certificates.
+        type: bool
+        default: false
+      force:
+        description:
+          - Overwrite existing custom or ACME certificate files.
+        type: bool
+        default: false
+  dns:
+    description:
+      - Manages the resolving DNS options of the node.
+    type: dict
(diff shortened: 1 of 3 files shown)


gyptazy:
Note: Unit tests will follow

plugins/modules/proxmox_node.py
```diff
@@ -0,0 +1,442 @@
...
+    description:
+      - The targeted node to perform actions on.
+    type: str
+    required: true
+  power_state:
+    description:
+      - Manages the power state of the node.
+    type: str
+    required: false
+    choices: ["present", "absent"]
```
Thulium-Drake: I'm not sure what this API does (does it do something with fencing agents?), but if it switches off the node, isn't it better to name them `online` and `shutdown` or something similar?

plugins/modules/proxmox_node.py
```diff
@@ -0,0 +1,442 @@
...
+        required_together=[('api_token_id', 'api_token_secret')],
+        supports_check_mode=True,
+    )
+
+    # Initialize objects and avoid re-polling the current
+    # nodes in the cluster in each function call.
+    proxmox = ProxmoxNodeAnsible(module)
+    nodes = proxmox.get_nodes()
+    proxmox.validate_node_name(nodes)
+    result = {"changed": False}
```
Thulium-Drake, line 415: As a more general note, I think we should standardize all module exits at some point to a similar structure like this (I also don't do it everywhere like this). But I think I prefer this style as it makes the rest of the code simpler to read.

plugins/modules/proxmox_node.py
gyptazy: I'm also fine with that naming. In the end, it's powering off / on (WOL) the node. I think having a power_state present (online) or absent (offline) is just about naming. I'm fine with both - if you prefer online/offline I'll adjust it that way :)

Please let me know! I'll adjust it with the upcoming unit test file (probably by the weekend or beginning of the next week - depending on my spare time).

plugins/modules/proxmox_node.py
gyptazy: Happy to hear - I mostly tend to do it this way for modules that have multiple options and to summarize them in the output. As you said, much simpler to read :)

plugins/modules/proxmox_node.py
Thulium-Drake: It's a discussion of semantics, I agree ;-) but I think `online`/`offline` makes it more clear for the users what the effect will be. Absent could also mean that the node's power management is disabled/removed or something like that.

gyptazy: I'm fine with it, let's do it that way. May concern was only more about that online / offline sounds more from a network perspective rather than from a PSU one. However, iLO also shows (or showed?) PSUs as "online" - so I'm fine with it ;)

squashed && pushed

gyptazy:
> Note: Unit tests will follow

* Unit test has been added.
* Switched power state from present / absent to online / offline

Thulium-Drake:
Well CI is happy, the tests look good :-)

gyptazy:
Thanks for merging that quickly :)

Issue #138, opened 2025-07-22
proxmox_cluster_ha_groups idempotency


rekup:
##### SUMMARY
While using the `proxmox_cluster_ha_groups` module I noticed that the module reports unexpected changes. While troubleshooting I saw that the `nodes` value returned by the proxmox api is not sorted and tha value is also not sorted in the module resulting in unexpected changes. Example API requests:

```
GET https://192.168.10.10:8006/api2/json/cluster/ha/groups
{
  "data": [
    {
      "nodes": "pve:20,pve2:20,pve3:20",
      "group": "all",
      "digest": "96a67da0731c0585243928441b3b71bbff0fa41d",
      "nofailback": 0,
      "comment": "HA group that enables a resource to run on all cluster nodes",
      "restricted": 0,
      "type": "group"
    }
  ]
}
```

Second request:

```
GET https://192.168.10.10:8006/api2/json/cluster/ha/groups
{
  "data": [
    {
      "comment": "HA group that enables a resource to run on all cluster nodes",
      "restricted": 0,
      "nofailback": 0,
      "group": "all",
      "digest": "96a67da0731c0585243928441b3b71bbff0fa41d",
      "nodes": "pve:20,pve3:20,pve2:20",
      "type": "group"
    }
  ]
}
```

The ordner of the `nodes` is different in the second request

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
proxmox_cluster_ha_groups

##### ANSIBLE VERSION

```paste below
ansible [core 2.16.14]
  config file = None
  configured module search path = ['/home/user/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/user/.local/share/virtualenvs/ansible-puzzle-OHOPjSlL/lib/python3.13/site-packages/ansible
  ansible collection location = /home/user/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/user/.local/share/virtualenvs/ansible-puzzle-OHOPjSlL/bin/ansible
  python version = 3.13.5 (main, Jun 12 2025, 00:00:00) [GCC 14.3.1 20250523 (Red Hat 14.3.1-1)] (/home/user/.local/share/virtualenvs/ansible-puzzle-OHOPjSlL/bin/python)
  jinja version = 3.1.6
  libyaml = True

```

##### COLLECTION VERSION

```paste below
Collection        Version
----------------- -------
community.proxmox 1.2.0  
```

[...]

PR #139, opened 2025-07-22
fix idempotency in proxmox_cluster_ha_groups module

rekup:
##### SUMMARY
This PR fixes the idempotency in the `proxmox_cluster_ha_groups` module.

Fixes https://github.com/ansible-collections/community.proxmox/issues/138

To fix it, I changed the param type from `str` to `list` and then sort the list provided by the user and the list provided by the api before comparing the two.

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION
Users should still be able to pass the nodes as string, it seems ansible automatically converts the string into a list. I tested this using these changes:

```diff
diff --git a/plugins/modules/proxmox_cluster_ha_groups.py b/plugins/modules/proxmox_cluster_ha_groups.py
index e0077bb..91eec8c 100644
--- a/plugins/modules/proxmox_cluster_ha_groups.py
+++ b/plugins/modules/proxmox_cluster_ha_groups.py
@@ -179,6 +179,10 @@ def run_module():
         supports_check_mode=False
     )
 
+    import q
+    q("User supplied nodes")
+    q(module.params['nodes'])
+
     proxmox = ProxmoxClusterHAGroupsAnsible(module)
 
     name = module.params['name']
```

And this task:

```
  - name: Create HA group for all cluster nodes
    community.proxmox.proxmox_cluster_ha_groups:
      api_host: "{{ proxmox_api_host }}"
      api_user: "{{ proxmox_api_user }}"
      api_token_id: "{{ proxmox_api_token_id }}"
      api_token_secret: "{{ proxmox_api_token_secret }}"
      state: "present"
      name: all
      comment: HA group that enables a resource to run on all cluster nodes
      nodes: pve:11,pve2:10,pve3:10 # <- passed as str
      nofailback: false
      restricted: false
    delegate_to: localhost
    run_once: true
```

This is the result:

```
 0.0s run_module: "User supplied nodes"='User supplied nodes'
 0.0s run_module: module.params['nodes']=['pve:11', 'pve2:10', 'pve3:10']
```

After fixing the idempotency I noted that the module did not report `changed` if a new group was first created. This seems to be due to a missing `return True` after the `POST` request which creates the HA group

### DIFF
--- a/plugins/modules/proxmox_cluster_ha_groups.py
+++ b/plugins/modules/proxmox_cluster_ha_groups.py
@@ -43,7 +43,8 @@ options:
         description: |
             List of cluster node members, where a priority can be given to each node. A resource bound to a group will run on the available nodes with the
             highest priority. If there are more nodes in the highest priority class, the services will get distributed to those nodes. The priorities have a
-            relative meaning only. The higher the number, the higher the priority. It can either be a string O(node_name:priority,node_name:priority) or an actual list of strings.
+            relative meaning only. The higher the number, the higher the priority.
+            It can either be a string C(node_name:priority,node_name:priority) or an actual list of strings.
         required: false
         type: list
         elements: str
@@ -129,7 +130,7 @@ class ProxmoxClusterHAGroupsAnsible(ProxmoxAnsible):
                 continue
 
             group["nodes"] = sorted(
-                    group.get("nodes", "").split(",")
+                group.get("nodes", "").split(",")
             )
 
             if (
--- a/plugins/modules/proxmox_cluster_ha_groups.py
+++ b/plugins/modules/proxmox_cluster_ha_groups.py
@@ -43,9 +43,10 @@ options:
         description: |
             List of cluster node members, where a priority can be given to each node. A resource bound to a group will run on the available nodes with the
             highest priority. If there are more nodes in the highest priority class, the services will get distributed to those nodes. The priorities have a
-            relative meaning only. The higher the number, the higher the priority.
+            relative meaning only. The higher the number, the higher the priority. It can either be a string O(node_name:priority,node_name:priority) or an actual list of strings.
         required: false
-        type: str
+        type: list
+        elements: str
     nofailback:
         description: |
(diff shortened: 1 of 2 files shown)


Thulium-Drake:
@commonism Do you want to take a look at these? :-)

plugins/modules/proxmox_cluster_ha_groups.py
```diff
@@ -127,6 +129,10 @@ def create(self, groups, name, comment, nodes, nofailback, restricted):
             if group["group"] != name:
                 continue
 
+            group["nodes"] = sorted(
+                group.get("nodes", "").split(",")
+            )
+
```
commonism, line 135: I'd stay with string, easier to work with in jinja2 than lists.

And therefore propose:

```python
            if (
                group.get("comment", ""),
                sorted(group.get("nodes", "").split(",")),
                bool(group.get("nofailback", 0)),
                bool(group.get("restricted", 0))
            ) == (comment, sorted(nodes.split(",")), nofailback, restricted):
```

rekup: Moving the splitting and sorting of `group.get("nodes", "")` to the condition makes sense, I'll change that.

Just to make sure, you propose to revert the `nodes` parameter from `list` to `str` as well? Because if we keep it as list it's up to the user to decide whether to use

```yaml
nodes: node1:10,node2:11
```

or

```yaml
nodes:
  - node1:10
  - node2:11
```

The `str` will still work even with `list` as param type (as per the `ADDITIONAL INFORMATION` section in this PR)

commonism:
LGTM

plugins/modules/proxmox_cluster_ha_groups.py
commonism: That said - LGTM.

commonism (approved):

Thulium-Drake:
@rekup Thanks for your work!
@commonism Thanks for reviewing!

PR #140, opened 2025-07-24
Added parameter master_api_password

### LINKED ISSUE #166: Using separate credentials while joining new node to cluster
Hello!

It seems the password using for joining the node to cluster is the very same password using for API connection:

https://github.com/ansible-collections/community.proxmox/blob/a8d3e59f9f856df318b0df2b02476777c8784797/plugins/modules/proxmox_cluster.py#L151

Could it be more optimal to use separate credentials here in case the existing cluster's root password and the joining node are not the same?

camillo:
##### SUMMARY
Introduce new, optional parameter **master_api_password** to proxmox_cluster module.
It is used for authentication against the master node.
If not set, api_password is used as before.

Fixes #166 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
proxmox_cluster

##### ADDITIONAL INFORMATION
Current implementation requires joining node and master node, sharing the same api_password.

### DIFF
--- a/changelogs/fragments/140-add-master-api-password.yml
+++ b/changelogs/fragments/140-add-master-api-password.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - proxmox_cluster -  Add master_api_password for authentication against master node (https://github.com/ansible-collections/community.proxmox/pull/140).
--- a/plugins/modules/proxmox_cluster.py
+++ b/plugins/modules/proxmox_cluster.py
@@ -84,6 +84,16 @@ EXAMPLES = r"""
     master_ip: "{{ primary_node }}"
     fingerprint: "{{ cluster_fingerprint }}"
     cluster_name: "devcluster"
+
+- name: Join a Proxmox VE Cluster with different API password
+    community.proxmox.proxmox_cluster:
+    api_host: proxmoxhost
+    api_user: root@pam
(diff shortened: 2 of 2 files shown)


Thulium-Drake:
Hi there,

Thanks for your work!

Can you elaborate a bit on why you need this, I don't fully follow when you'd need this. And can you also add an example to the docs for that situation?

Thanks!

camillo:
Hiho!

Thanks for looking into it.
I added an example.

Scenario is:
- Two fresh installed proxmox server
- Both have a token for root@pam configured
- They have different root passwords

New Proxmox cluster is created on server1

Joining process for server2 is:
1. use api_user & api_password to connect to server2
2. proxmox_cluster.py, calls server1's api for joining the cluster
    ``` python
                self.proxmox_api.cluster.config.join.post(
                    hostname=master_ip,
                    fingerprint=fingerprint,
                    password=api_password
                )
    ```

The call to server1 use the same password that was used for authenticate against server2. If they do not match, the join.post(…) fails with an authentication error.

Thulium-Drake:
Ah, right, thanks for the clarification, it never occurred to me that not all root accounts are necessarily the same :sweat_smile:

This makes sense to me now! And it LGTM.

Can you address the issues mentioned by the Sanity tests?

PR #143, opened 2025-07-27
Pass template to fetch_template and fetch_and_verify

### LINKED ISSUE #142: proxmox_template does not respect template parameter for URLs
##### SUMMARY
When using the `url` parameter for `proxmox_template`, the `template` parameter is ignored and the name of the template is derived from the URL which may not always be correct/desired. This also goes against the description of the parameter which states it's the **_name_** of the template.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
proxmox_template

##### ANSIBLE VERSION
```
ansible [core 2.18.7]
  config file = None
  configured module search path = ['/home/danthonywalker/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/danthonywalker/Code/ansible/.venv/lib64/python3.13/site-packages/ansible
  ansible collection location = /home/danthonywalker/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/danthonywalker/Code/ansible/.venv/bin/ansible
  python version = 3.13.5 (main, Jun 12 2025, 00:00:00) [GCC 15.1.1 20250521 (Red Hat 15.1.1-2)] (/home/danthonywalker/Code/ansible/.venv/bin/python)
  jinja version = 3.1.6
  libyaml = True
```

##### COLLECTION VERSION
```
Collection        Version
----------------- -------
community.proxmox 1.1.0
```

##### CONFIGURATION
```
CONFIG_FILE() = None
EDITOR(env: EDITOR) = /usr/bin/nano

GALAXY_SERVERS:
```

##### OS / ENVIRONMENT
I am running Fedora, however, I do not believe this issue is OS-specific to the Ansible control node.

##### STEPS TO REPRODUCE
Consider the following task with required parameters omitted for brevity:
```yaml
- name: Download
  community.proxmox.proxmox_template:
    url: "{{ iso_url }}"
    template: "{{ iso_name }}"
    content_type: iso
    state: present
```

The name saved to storage will be the basename of the URL, which is sometimes not desired. For example, I am trying to download `https://factory.talos.dev/image/376567988ad370138ad8b2698212367b8edcb69b5fd68c80be1f2ec7d603b4ba/v1.10.5/nocloud-amd64.iso` which results in a template name of `nocloud-amd64.iso` which is too generic and can collide to other versions of the ISO. I wanted to pass in a custom name of `talos-376567988ad370138ad8b2698212367b8edcb69b5fd68c80be1f2ec7d603b4ba-1.10.5-nocloud-amd64.iso` however no matter what you set for `template` it is ignored.

##### EXPECTED RESULTS
I expect `template` to take precedence on the name of the template when it is provided as stated by its documentation. This is also possible through the Proxmox UI.

##### ACTUAL RESULTS
If `url` is passed, then `template` is ignored and the basename of the URL is always used.

##### ACTUAL RESULTS
This is likely caused by these lines not considering the `module.params['template']`:
https://github.com/ansible-collections/community.proxmox/blob/main/plugins/modules/proxmox_template.py#L329-L350

In fact `fetch_template` assumes the filename on its own, when this could be passed as a parameter:
https://github.com/ansible-collections/community.proxmox/blob/88e497ca5506adc178832167e46c9dd9ed2c7091/plugins/modules/proxmox_template.py#L233

danthonywalker:
##### SUMMARY
This passes `module.params['template']` to `fetch_template` and `fetch_and_verify` in `proxmox_template`. This fixes #142. If the template is not provided, it falls back to the defaults it had previously (effectively...technically it's changing from `os.path.basename(url)` to `os.path.basename(urlparse(url).path)` but that is semantically equivalent).

Note that this does not fix `src`. The main reason being is I'm unsure how to pass a template name to the API call for uploading. It's probably trivial to followup on though.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
`proxmox_template`

### DIFF
--- a/plugins/modules/proxmox_template.py
+++ b/plugins/modules/proxmox_template.py
@@ -225,12 +225,12 @@ class ProxmoxTemplateAnsible(ProxmoxAnsible):
         except Exception as e:
             self.module.fail_json(msg="Uploading template %s failed with error: %s" % (realpath, e))
 
-    def fetch_template(self, node, storage, content_type, url, timeout):
+    def fetch_template(self, node, storage, content_type, url, timeout, template):
         """Fetch a template from a web url source using the proxmox download-url endpoint
         """
         try:
             taskid = self.proxmox_api.nodes(node).storage(storage)("download-url").post(
-                url=url, content=content_type, filename=os.path.basename(url)
+                url=url, content=content_type, filename=template
             )
             return self.task_status(node, taskid, timeout)
         except Exception as e:
@@ -256,13 +256,13 @@ class ProxmoxTemplateAnsible(ProxmoxAnsible):
             time.sleep(1)
         return False
 
-    def fetch_and_verify(self, node, storage, url, content_type, timeout, checksum, checksum_algorithm):
+    def fetch_and_verify(self, node, storage, url, content_type, timeout, checksum, checksum_algorithm, template):
         """ Fetch a template from a web url, then verify it using a checksum.
         """
         data = {
             'url': url,
             'content': content_type,
-            'filename': os.path.basename(url),
+            'filename': template,
             'checksum': checksum,
             'checksum-algorithm': checksum_algorithm}
         try:
@@ -310,11 +310,10 @@ def main():
         content_type = module.params['content_type']
         src = module.params['src']
         url = module.params['url']
+        template = module.params['template']
 
         # download appliance template
         if content_type == 'vztmpl' and not (src or url):
-            template = module.params['template']
-
             if not template:
                 module.fail_json(msg='template param for downloading appliance template is mandatory')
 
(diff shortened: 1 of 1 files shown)


Thulium-Drake:
Thanks for your work, I've approved the checks, can you address the issues it's reporting? :-)

danthonywalker:
It seems like the only issue was indentation, which I fixed. The unit tests seem broken prior to this change since it's referencing completely unrelated modules.

evarring:
Just ran into the same issue today as i am downloading templates from linuxcontainers dot org and they are named rootfs.tar.xz by default (even if i set the template value to something like linuxcontainer.tar.xz). I tested out your commit and the "template" value is now respected and the image saved with the correct name i set :)

Thank you!

Thulium-Drake:
!recheck

Thulium-Drake:
Thanks for the fix! :-)

Issue #149, opened 2025-08-01
feature: Add new module proxmox_storage


gyptazy:
##### SUMMARY
We often create multiple clusters with shared storage and I'd like to do it in a proper way aligned to all other roles that ensures that the API including all authentications and authorizations layers is used. Therefore, an Ansible module for managing storage (especially adding/removing) should be added. This module should be called `proxmox_storage`, next to the already present `$name_info` modules.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
proxmox_storage

##### ADDITIONAL INFORMATION
I'll provide a first draft asap. While this draft is working, we should discuss the layout of the module args (tbd within the PR).

PR #150, opened 2025-08-01
feature: Add new module proxmox_storage

gyptazy:
feature: Add new module proxmox_storage

##### SUMMARY
Allow adding/removing of storage(s) in Proxmox VE nodes and clusters.

With the given module storage(s) can be added to Proxmox VE nodes and clusters.
This provides the ability to easily integrate (shared) storages like:
* NFS
* CIFS/SMB
* Proxmox Backup Server
* iSCSI

Fixes: #149
Sponsored-by: credativ GmbH

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
proxmox_storage

##### ADDITIONAL INFORMATION
* We need to discuss the integration of this module! Do not merge yet!
* Unit tests will be added afterwards

### DIFF
--- a/plugins/modules/proxmox_storage.py
+++ b/plugins/modules/proxmox_storage.py
@@ -1,0 +1,501 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2025, Florian Paul Azim Hoberg (@gyptazy) <florian.hoberg@credativ.de>
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = r"""
+module: proxmox_storage
+version_added: 1.3.0
+short_description: Manage storage in PVE clusters and nodes
+description:
+  - Manage storage in PVE clusters and nodes.
+author: Florian Paul Azim Hoberg (@gyptazy)
+attributes:
+  check_mode:
+    support: full
+  diff_mode:
+    support: none
+options:
+  nodes:
+    description:
+      - A list of Proxmox VE nodes on which the target storage is enabled.
+      - Required when C(state=present).
+    type: list
+    elements: str
+    required: false
+  name:
+    description:
+      - The name of the storage displayed in the storage list.
+    type: str
+    required: true
+  state:
+    description:
+      - The state of the defined storage type to perform.
+    choices: ["present", "absent"]
+    type: str
+  type:
+    description:
+      - The storage type/protocol to use when adding the storage.
+    type: str
+    required: true
+    choices: ['cephfs', 'cifs', 'iscsi', 'nfs', 'pbs']
+  cephfs_options:
+    description:
+      - Extended information for adding CephFS storage.
+    type: dict
+    suboptions:
+      monhost:
+        description:
+          - The hostname or IP address of the monhost.
+        type: list
+        elements: str
+        required: false
+      username:
+        description:
+          - The required username for the storage system.
+        type: str
+        required: false
+      password:
+        description:
+          - The required password for the storage system.
+        type: str
+        required: false
+      path:
+        description:
+          - The path to be used within the CephFS.
+        type: str
+        default: '/'
+        required: false
+      subdir:
+        description:
+          - The subdir to be used within the CephFS.
+        type: str
+        required: false
+      client_keyring:
+        description:
+          - The client keyring to be used.
+        type: str
+        required: false
+      fs_name:
+        description:
+          - The Ceph filesystem name
+        type: str
+        required: false
+  cifs_options:
+    description:
+      - Extended information for adding CIFS storage.
+    type: dict
+    suboptions:
+      server:
+        description:
+          - The hostname or IP address of the remote storage system.
+        type: str
+        required: false
+      username:
+        description:
+          - The required username for the storage system.
+        type: str
+        required: false
+      password:
+        description:
+          - The required password for the storage system.
+        type: str
+        required: false
+      share:
+        description:
+          - The share to be used from the remote storage system.
+        type: str
+        required: false
+      domain:
+        description:
+          - The required domain for the CIFS share.
+        type: str
+        required: false
+      smb_version:
+        description:
+          - The minimum SMB version to use for.
+        type: str
+        required: false
+  nfs_options:
+    description:
+      - Extended information for adding NFS storage.
+    type: dict
+    suboptions:
+      server:
+        description:
+          - The hostname or IP address of the remote storage system.
+        type: str
+        required: false
+      export:
+        description:
+          - The required NFS export path.
+        type: str
+        required: false
+      options:
+        description:
+          - Additional NFS related mount options (e.g., version, pNFS).
+        type: str
+        required: false
+  iscsi_options:
+    description:
+      - Extended information for adding iSCSI storage.
+    type: dict
+    suboptions:
+      portal:
+        description:
+          - The hostname or IP address of the remote storage system as the portal address.
+        type: str
+        required: false
+      target:
+        description:
+          - The required iSCSI target.
+        type: str
+        required: false
+  pbs_options:
+    description:
+      - Extended information for adding Proxmox Backup Server as storage.
+    type: dict
+    suboptions:
+      server:
+        description:
+          - The hostname or IP address of the Proxmox Backup Server.
+        type: str
+        required: false
+      username:
+        description:
+          - The required username for the Proxmox Backup Server.
+        type: str
+        required: false
+      password:
+        description:
+          - The required password for the Proxmox Backup Server.
+        type: str
+        required: false
+      datastore:
+        description:
+          - The required datastore to use from the Proxmox Backup Server.
+        type: str
+        required: false
+      fingerprint:
+        description:
+          - The required fingerprint of the Proxmox Backup Server system.
+        type: str
+        required: false
+  content:
+    description:
+      - The desired content that should be used with this storage type.
+      - Required when C(state=present).
+    type: list
+    required: false
+    elements: str
+    choices: ["images", "snippets", "import", "iso", "backup", "rootdir", "vztmpl"]
+extends_documentation_fragment:
+  - community.proxmox.proxmox.actiongroup_proxmox
+  - community.proxmox.proxmox.documentation
+  - community.proxmox.attributes
+"""
+
+EXAMPLES = r"""
+- name: Add PBS storage to Proxmox VE Cluster
+  community.proxmox.proxmox_storage:
+    api_host: proxmoxhost
+    api_user: root@pam
+    api_password: password123
+    validate_certs: false
+    nodes: ["de-cgn01-virt01", "de-cgn01-virt02"]
+    state: present
+    name: backup-backupserver01
+    type: pbs
+    pbs_options:
+      server: proxmox-backup-server.example.com
+      username: backup@pbs
+      password: password123
+      datastore: backup
+      fingerprint: "F3:04:D2:C1:33:B7:35:B9:88:D8:7A:24:85:21:DC:75:EE:7C:A5:2A:55:2D:99:38:6B:48:5E:CA:0D:E3:FE:66"
+      export: "/mnt/storage01/b01pbs01"
+    content: ["backup"]
+- name: Add NFS storage to Proxmox VE Cluster
+  community.proxmox.proxmox_storage:
+    api_host: proxmoxhost
+    api_user: root@pam
+    api_password: password123
+    validate_certs: false
+    nodes: ["de-cgn01-virt01", "de-cgn01-virt02"]
+    state: present
+    name: net-nfsshare01
+    type: nfs
+    nfs_options:
+      server: 10.10.10.94
+      export: "/mnt/storage01/s01nfs01"
+    content: ["rootdir", "images"]
+- name: Add iSCSI storage to Proxmox VE Cluster
+  community.proxmox.proxmox_storage:
+    api_host: proxmoxhost
+    api_user: root@pam
+    api_password: password123
+    validate_certs: false
+    nodes: ["de-cgn01-virt01", "de-cgn01-virt02", "de-cgn01-virt03"]
+    state: present
+    type: iscsi
+    name: net-iscsi01
+    iscsi_options:
+      portal: 10.10.10.94
+      target: "iqn.2005-10.org.freenas.ctl:s01-isci01"
+    content: ["rootdir", "images"]
+- name: Remove storage from Proxmox VE Cluster
+  community.proxmox.proxmox_storage:
+    api_host: proxmoxhost
+    api_user: root@pam
+    api_password: password123
+    validate_certs: false
+    state: absent
+    name: net-nfsshare01
+    type: nfs
+"""
+
+RETURN = r"""
+storage:
+  description: Status message about the storage action.
+  returned: success
+  type: str
+  sample: "Storage 'net-nfsshare01' created successfully."
+"""
+
+from ansible.module_utils.basic import AnsibleModule
+from ansible_collections.community.proxmox.plugins.module_utils.proxmox import (
+    proxmox_auth_argument_spec, ProxmoxAnsible)
+
+
+class ProxmoxNodeAnsible(ProxmoxAnsible):
+    def add_storage(self):
+        changed = False
+        result = "Unchanged"
+        storage_name = self.module.params['name']
+        storage_type = self.module.params['type']
+        nodes = self.module.params['nodes']
+        content = self.module.params.get('content')
+
+        # Create payload for storage creation
+        payload = {
+            'storage': storage_name,
+            'type': storage_type,
+            'nodes': nodes,
+            'content': content
+        }
+
+        # Validate required parameters based on storage type
+        if storage_type == "cephfs":
+            cephfs_options = self.module.params.get(f'{storage_type}_options', {})
+            monhost = cephfs_options.get('monhost', '')
+            username = cephfs_options.get('username')
+            password = cephfs_options.get('password')
+            path = cephfs_options.get('path', '/')
+            subdir = cephfs_options.get('subdir', None)
+            client_keyring = cephfs_options.get('client_keyring')
+            fs_name = cephfs_options.get('fs_name')
+
+            if not monhost == "":
+                payload['monhost'] = monhost
+            if username:
+                payload['username'] = username
+            if password:
+                payload['password'] = password
+            payload['path'] = path
+            payload['subdir'] = subdir
+            if client_keyring:
+                payload['client_keyring'] = client_keyring
+            if fs_name:
+                payload['fs_name'] = fs_name
+
+        if storage_type == "cifs":
+            cifs_options = self.module.params.get(f'{storage_type}_options', {})
+            server = cifs_options.get('server')
+            share = cifs_options.get('share')
+            if not all([server, share]):
+                self.module.fail_json(msg="CIFS storage requires 'server' and 'share' parameters.")
+            else:
+                payload['server'] = server
+                payload['share'] = share
+
+        if storage_type == "iscsi":
+            iscsi_options = self.module.params.get(f'{storage_type}_options', {})
+            portal = iscsi_options.get('portal')
+            target = iscsi_options.get('target')
+            if not all([portal, target]):
+                self.module.fail_json(msg="iSCSI storage requires 'portal' and 'target' parameters.")
+            else:
+                payload['portal'] = portal
+                payload['target'] = target
+
+        if storage_type == "nfs":
+            nfs_options = self.module.params.get(f'{storage_type}_options', {})
+            server = nfs_options.get('server')
+            export = nfs_options.get('export')
+            if not all([server, export]):
+                self.module.fail_json(msg="NFS storage requires 'server' and 'export' parameters.")
+            else:
+                payload['server'] = server
+                payload['export'] = export
+
+        if storage_type == "pbs":
+            pbs_options = self.module.params.get(f'{storage_type}_options', {})
+            server = pbs_options.get('server')
+            username = pbs_options.get('username')
+            password = pbs_options.get('password')
+            datastore = pbs_options.get('datastore')
+            fingerprint = pbs_options.get('fingerprint')
+            if not all([server, datastore, username, password]):
+                self.module.fail_json(msg="PBS storage requires 'server', 'username', 'password', 'datastore' and 'fingerprint' parameters.")
+
+        # Check Mode validation
+        if self.module.check_mode:
+            try:
+                existing_storages = self.proxmox_api.storage.get()
+            except Exception as e:
+                self.module.fail_json(msg=f"Failed to retrieve storage list: {e}")
+
+            for storage in existing_storages:
+                if storage.get("storage") == storage_name:
+                    changed = False
+                    function_result = f"Storage '{storage_name}' already present."
+                    result = {"changed": changed, "msg": function_result}
+                    self.module.exit_json(**result)
+
+                changed = True
+                function_result = f"Storage '{storage_name}' would be created."
+                result = {"changed": changed, "msg": function_result}
+                self.module.exit_json(**result)
+
+        # Add storage
+        try:
+            self.proxmox_api.storage.post(**payload)
+            changed = True
+            result = f"Storage '{storage_name}' created successfully."
+        except Exception as e:
+            error_msg = str(e)
+            if "already defined" in error_msg:
+                changed = False
+                result = f"Storage '{storage_name}' already present."
+            else:
+                self.module.fail_json(msg=f"Failed to create storage: {error_msg}")
+
+        return changed, result
+
+    def remove_storage(self):
+        changed = False
+        result = "Unchanged"
+        storage_name = self.module.params["name"]
+
+        # Check Mode validation
+        if self.module.check_mode:
+            try:
+                existing_storages = self.proxmox_api.storage.get()
+            except Exception as e:
+                self.module.fail_json(msg=f"Failed to retrieve storage list: {e}")
+
+            for storage in existing_storages:
+                if storage.get("storage") == storage_name:
+                    changed = True
+                    result = {"changed": changed, "msg": f"Storage '{storage_name}' would be deleted."}
+                    self.module.exit_json(**result)
+
+            changed = False
+            result = {"changed": changed, "msg": f"Storage '{storage_name}' does not exist."}
+            self.module.exit_json(**result)
+
+        # Remove storage
+        try:
+            existing_storages = self.proxmox_api.storage.get()
+            if not any(s.get("storage") == storage_name for s in existing_storages):
+                changed = False
+                result = f"Storage '{storage_name}' does not exist."
+                return changed, result
+
+            self.proxmox_api.storage(storage_name).delete()
+            changed = True
+            result = f"Storage '{storage_name}' removed successfully."
+
+        except Exception as e:
+            self.module.fail_json(msg=f"Failed to delete storage '{storage_name}': {e}")
+
+        return changed, result
+
+
+def main():
+    module_args = proxmox_auth_argument_spec()
+
+    storage_args = dict(
+        nodes=dict(type='list', elements='str',),
+        name=dict(type='str', required=True),
+        state=dict(choices=['present', 'absent']),
+        type=dict(choices=['cephfs', 'cifs', 'iscsi', 'nfs', 'pbs'], required=True),
+        cephfs_options=dict(type='dict', options={
+            'monhost': dict(type='list', elements='str'),
+            'username': dict(type='str'),
+            'password': dict(type='str', no_log=True),
+            'path': dict(type='str', default='/'),
+            'subdir': dict(type='str',),
+            'fs_name': dict(type='str',),
+            'client_keyring': dict(type='str', no_log=True)
+        }),
+        cifs_options=dict(type='dict', options={
+            'server': dict(type='str'),
+            'username': dict(type='str'),
+            'password': dict(type='str', no_log=True),
+            'share': dict(type='str'),
+            'domain': dict(type='str'),
+            'smb_version': dict(type='str')
+        }),
+        nfs_options=dict(type='dict', options={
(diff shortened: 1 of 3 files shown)


gyptazy:
Hey everyone, hey @Thulium-Drake, hey @felixfontein !

I'm not quite sure about the current arg spec design of this module and like to discuss this before proceeding.
Currently, it is designed this way:

``` 
- name: Add NFS storage to Proxmox VE Cluster
  community.proxmox.proxmox_storage:
    api_host: proxmoxhost
    api_user: root@pam
    api_password: password123
    validate_certs: false
    node_name: de-cgn01-virt01
    nodes: ["de-cgn01-virt01", "de-cgn01-virt02"]
    state: present
    name: net-nfsshare01
    type: nfs
    server: 10.10.10.94
    export: "/mnt/storage01/s01nfs01"
    content: ["rootdir", "images"]
- name: Add iSCSI storage to Proxmox VE Cluster
  community.proxmox.proxmox_storage:
    api_host: proxmoxhost
    api_user: root@pam
    api_password: password123
    validate_certs: false
    node_name: de-cgn01-virt01
    nodes: ["de-cgn01-virt01", "de-cgn01-virt02", "de-cgn01-virt03"]
    state: present
    name: net-iscsi01
    type: iscsi
    portal: 10.10.10.94
    target: "iqn.2005-10.org.freenas.ctl:s01-isci01"
    content: ["rootdir", "images"]
```

As we can see here, we have for NFS:
```
    type: nfs
    server: 10.10.10.94
    export: "/mnt/storage01/s01nfs01"
```

and for iSCSI:
```
    type: iscsi
    portal: 10.10.10.94
    target: "iqn.2005-10.org.freenas.ctl:s01-isci01"
```

this means we have more or less the same meaning behind the things but use them in the defined scope of usage (server = portal, export = target) and this is totally fine.

However, this results in many top level keys that can be defined (also also often cannot be combined). Like `server` and `portal` but also several other ones (nfs: export, cifs: share, domain, iscsi: portal, target, ...). 

The current idea is to validate them within the module for proper usage and allowed combinations (see also https://github.com/ansible-collections/community.proxmox/pull/150/files#diff-61372054df16d605a24c34b3dfd2b526dd071efbf30d5c6213c67aefd03074d3R217). This decision has been taken to work as near as possible to the Proxmox API and to use the same key/value design (compare: https://pve.proxmox.com/pve-docs/api-viewer/#/storage/{storage} PUT Methods). Within the current implementation, we would match the API but I'm not quite sure if this is the preferred way for Ansible users (which might be a bit confusing or maybe even more logical - completely depends on the point of view).

Another possibility - which would be a bite cleaner from an Ansbile user perspective - could lead into a following syntax:

```
- name: Add NFS storage to Proxmox VE Cluster
  community.proxmox.proxmox_storage:
    api_host: proxmoxhost
    api_user: root@pam
    api_password: password123
    validate_certs: false
    node_name: de-cgn01-virt01
    nodes: ["de-cgn01-virt01", "de-cgn01-virt02"]
    state: present
    name: net-nfsshare01
    type: nfs
    storage_params:
        server: 10.10.10.94
        export: "/mnt/storage01/s01nfs01"
        content: ["rootdir", "images"]
- name: Add iSCSI storage to Proxmox VE Cluster
  community.proxmox.proxmox_storage:
    api_host: proxmoxhost
    api_user: root@pam
    api_password: password123
    validate_certs: false
    node_name: de-cgn01-virt01
    nodes: ["de-cgn01-virt01", "de-cgn01-virt02", "de-cgn01-virt03"]
    state: present
    name: net-iscsi01
    type: iscsi
    storage_params:
        portal: 10.10.10.94
        target: "iqn.2005-10.org.freenas.ctl:s01-isci01"
        content: ["rootdir", "images"]
```

I'm open for further feedback :)

Thanks,
gyptazy

Thulium-Drake:
Hi @gyptazy, thanks for this wonderful module! (it comes just as we're working towards fully automating a cluster for a client ;-) )

I think I'm in favor of having the `storage_params` version, it's cleaner and I _think_ it's easier to validate each usecase on it's own as you'd only need to validate whether a `type: 'nfs'` is actually valid NFS, and you don't have to care about the rest.

Also, while you're at it, can you have a look on how to integrate CephFS as well as RBD storages? 

I have marked the PR with 'WIP:', this will disable the merge logic, but still allow discussion and tests

gyptazy:
Hey @Thulium-Drake,

thanks for your quick reply!

> [...] thanks for this wonderful module! (it comes just as we're working towards fully automating a cluster for a client ;-) )

Same here, including some additional modules in the pipeline that will be pushed soon (firewall, sdn, etc.) :)

> I think I'm in favor of having the `storage_params` version, it's cleaner and I _think_ it's easier to validate each usecase on it's own as you'd only need to validate whether a `type: 'nfs'` is actually valid NFS, and you don't have to care about the rest.

I'm fine and open for this - that's why I asked for other opinions - appreciate your feedback :)

> Also, while you're at it, can you have a look on how to integrate CephFS as well as RBD storages?

We're not actively using CephFS and RBD. While I also thought about it, it would have been an "assumption" integration approach without further tests that it really works. Maybe I can spend two hours this weekend to give at least CephFS a quick try which should be pretty quick to set up in a dirty way for further integrations tests.

I'll push the adjustments asap. After merge, we can proceed with the next module :P 

Thanks a lot and have a nice weekend!

gyptazy:
@Thulium-Drake 

one thing just popped up in my mind... 

Generic as `storage_otps` (for all types just as sub options) like:
```
  storage_opts:
    description:
      - Extended storage specific information when adding to Proxmox VE Node(s) or Clusters
    type: dict
    suboptions:
      server: #This would be for NFS
        description:
          -  The hostname or IP address of the remote storage system.
        type: str
      portal: #This would be for iSCSI
        description:
          -  The hostname or IP address of the remote storage system.
        type: str
```

or directly scoped (for each storage type in `type`):
```
  nfs_opts:
    description:
      - Extended NFS specific information when adding to Proxmox VE Node(s) or Clusters
    type: dict
    suboptions:
        server: #This would be for NFS
        description:
          -  The hostname or IP address of the remote storage system.
        type: str
        export:
        description:
          - The required mount path of the storage on the Proxmox VE node(s).
        type: str
        required: true
```

I'd prefer the scoped ones which would look like:

```
options:
  node_name:
    description:
      - The targeted Proxmox VE node to perform actions on.
    type: str
    required: true
  nodes:
    description:
      - A list of Proxmox VE nodes where the storage action should be performed on.
    type: list
    required: true
  name:
    description:
      - The name of the storage displayed in the storage list.
    type: str
    required: true
  state:
    description:
      - The state of the defined storage type to perform.
    choices: ["present", "absent"]
    type: str
  type:
    description:
      - The storage type/protocol to use when adding the storage. 
    type: str
    required: true
    choices: ['cephfs', 'cifs', 'iscsi', 'nfs', 'pbs']

  cephfs_options:
    description:
      - Extended information for adding CephFS storage.
    type: dict
    suboptions:
      monhost:
        description:
          - The hostname or IP address of the monhost.
        type: str
        required: false
      username:
        description:
          - The required username for the storage system.
        type: str
        required: false
      password:
        description:
          - The required password for the storage system.
        type: str
        required: false
      path:
        description:
          - The path to be used within the CephFS.
        type: str
        required: false
      subdir:
        description:
          - The subdir to be used within the CephFS.
        type: str
        required: false
      client_keyring:
        description:
          - The client keyring to be used.
        type: str
        required: false

  cifs_options:
    description:
      - Extended information for adding CIFS storage.
    type: dict
    suboptions:
      server:
        description:
          - The hostname or IP address of the remote storage system.
        type: str
        required: false
      username:
        description:
          - The required username for the storage system.
        type: str
        required: false
      password:
        description:
          - The required password for the storage system.
        type: str
        required: false
      share:
        description:
          - The share to be used from the remote storage system.
        type: str
        required: false
      domain:
        description:
          - The required domain for the CIFS share.
        type: str
        required: false
      smb_version:
        description:
          - The minimum SMB version to use for.
        type: str
        required: false

  nfs_options:
    description:
      - Extended information for adding NFS storage.
    type: dict
    suboptions:
      server:
        description:
          - The hostname or IP address of the remote storage system.
        type: str
        required: false
      export:
        description:
          - The required NFS export path.
        type: str
        required: false
      options:
        description:
          - Additional NFS related mount options (e.g., version, pNFS).
        type: str
        required: false

  iscsi_options:
    description:
      - Extended information for adding iSCSI storage.
    type: dict
    suboptions:
      portal:
        description:
          - The hostname or IP address of the remote storage system as the portal address.
        type: str
        required: true
        required: false
      target:
        description:
          - The required iSCSI target.
        type: str
        required: false

  pbs_options:
    description:
      - Extended information for adding Proxmox Backup Server as storage.
    type: dict
    suboptions:
      server:
        description:
          - The hostname or IP address of the Proxmox Backup Server.
        type: str
        required: false
      username:
        description:
          - The required username for the Proxmox Backup Server.
        type: str
        required: false
      password:
        description:
          - The required password for the Proxmox Backup Server.
        type: str
        required: false
      datastore:
        description:
          - The required datastore to use from the Proxmox Backup Server.
        type: str
        required: false
      fingerprint:
        description:
          - The required fingerprint of the Proxmox Backup Server system.
        type: str
        required: false

  content:
    description:
      - The desired content that should be used with this storage type.
    type: list
    required: true
    choices: ["images", "snippets", "import", "iso", "backup", "rootdir", "vztmpl"]
extends_documentation_fragment:
  - community.proxmox.proxmox.actiongroup_proxmox
  - community.proxmox.proxmox.documentation
  - community.proxmox.attributes
"""

EXAMPLES = r"""
- name: Add PBS storage to Proxmox VE Cluster
  community.proxmox.proxmox_storage:
    api_host: proxmoxhost
    api_user: root@pam
    api_password: password123
    validate_certs: false
    node_name: de-cgn01-virt01
    nodes: ["de-cgn01-virt01", "de-cgn01-virt02"]
    state: present
    name: backup-backupserver01
    type: pbs
    pbs_options:
      server: proxmox-backup-server.example.com
      username: backup@pbs
      password: password123
      datastore: backup
      fingerprint: "F3:04:D2:C1:33:B7:35:B9:88:D8:7A:24:85:21:DC:75:EE:7C:A5:2A:55:2D:99:38:6B:48:5E:CA:0D:E3:FE:66"
      export: "/mnt/storage01/b01pbs01"
    content: ["backup"]
- name: Add NFS storage to Proxmox VE Cluster
  community.proxmox.proxmox_storage:
    api_host: proxmoxhost
    api_user: root@pam
    api_password: password123
    validate_certs: false
    node_name: de-cgn01-virt01
    nodes: ["de-cgn01-virt01", "de-cgn01-virt02"]
    state: present
    type: nfs
    nfs_otpions:
      name: net-nfsshare01
      server: 10.10.10.94
      export: "/mnt/storage01/s01nfs01"
    content: ["rootdir", "images"]
- name: Add iSCSI storage to Proxmox VE Cluster
  community.proxmox.proxmox_storage:
    api_host: proxmoxhost
    api_user: root@pam
    api_password: password123
    validate_certs: false
    node_name: de-cgn01-virt01
    nodes: ["de-cgn01-virt01", "de-cgn01-virt02", "de-cgn01-virt03"]
    state: present
    type: iscsi
    iscsi_options:
      name: net-iscsi01
      portal: 10.10.10.94
      target: "iqn.2005-10.org.freenas.ctl:s01-isci01"
    content: ["rootdir", "images"]
```

Cheers,
gyptazy

felixfontein:
If the options per type are so different, I think having a different `xxx_params` option per type would be best. (One example of a module which does this: https://docs.ansible.com/ansible/devel/collections/community/hrobot/boot_module.html#parameters - there I left `type` away and simply used the option key such as `install_linux` as the type indicator, instead of having something like `type: linux`. Whether leaving `type` away is a good idea is another question; it might be a bad idea if you need to disable specific types with `type: xxx`, `state: absent`.)

If one common suboption is used for all types, maybe it makes sense to use different names than Proxmox itself. But that also has quite some downsides, like sooner or later Proxmox will add a new option that has the same name than an option name you already picked or something else. Using the original Proxmox names (and having one option for everything with all different names in it) can so cause problems, if different storage types use the same name for very different objects (like one uses `foo` for a list of strings, the other uses `foo` for a boolean). That's why I prefer having different `xxx_params` options per type.

(`xxx_params` is a placeholder for however you want to name them. I don't have a preference about the actual name / naming pattern. I just had to use something in this comment ;) )

gyptazy:
Hey @Thulium-Drake, hey @felixfontein,

just pushed a new iteration including the discussed changes. Let me know if this fits to you - afterwards I'll polish it up and create the unit tests.

Have a nice weekend!
gyptazy

gyptazy:
* Integrate the discussed adjustments
* Added unit test

=> Ready for review

plugins/modules/proxmox_storage.py
```diff
@@ -0,0 +1,520 @@
...
+# Copyright (c) 2025, Florian Paul Azim Hoberg (@gyptazy) <florian.hoberg@credativ.de>
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = r"""
+module: proxmox_storage
+version_added: 1.2.0
```
Thulium-Drake:
```suggestion
version_added: 1.3.0
```

Thulium-Drake:
Awesome work! There's one case that comes to mind, and I think it will break right now.

Say you only want to update the contents for an existing storage. How would what be handled (e.g. some storage named 'data' on which you want to enable ISO storage)?

I'd be OK with the answer being 'define everything', but it might be something to think about :-)

plugins/modules/proxmox_storage.py
```diff
@@ -0,0 +1,520 @@
...
+    support: none
+options:
+  node_name:
+    description:
+      - The targeted Proxmox VE node to perform actions on.
+    type: str
+    required: true
+  nodes:
+    description:
+      - A list of Proxmox VE nodes where the storage action should be performed on.
```
Thulium-Drake:
```suggestion
      - A list of Proxmox VE nodes on which the target storage is enabled.
```

plugins/modules/proxmox_storage.py
```diff
@@ -0,0 +1,520 @@
...
+description:
+  - Manage storage in PVE clusters and nodes.
+author: Florian Paul Azim Hoberg (@gyptazy)
+attributes:
+  check_mode:
+    support: full
+  diff_mode:
+    support: none
+options:
+  node_name:
```
Thulium-Drake: Isn't this redundant to `api_host`?

gyptazy:
> Say you only want to update the contents for an existing storage. How would what be handled (e.g. some storage named 'data' on which you want to enable ISO storage)?
> I'd be OK with the answer being 'define everything', but it might be something to think about :-)

I also had this in mind, but unfortunately we have no patch methods within the Proxmox API for storage included. This means, that we get immediately an error returned, when the storage id (name) is already in use: `Failed to create storage: 500 Internal Server Error: create storage failed: storage ID 'net-nfsshare01' already defined"`. 

As a result, we have the following possibilities:
* Deal with it and keep it as it is. The ID is a unique constraint that can not be altered. Therefore, it must be intentionally removed and re-added when already present.
* We force remove the storage, update the values and add it within the module. It's not a big deal, but as an Ansible user I would not expect that updating a content parameter would remove me the whole storage from the cluster - which might even create a huge mess in production setups because of an unexpected event.
* We could update the content within the storage configuration but it would omit the API. I'd highly like to only use the API methods to make sure we do not need any SSH access to systems using those modules.

Happy to hear how you'd like to proceed.

plugins/modules/proxmox_storage.py
```diff
@@ -0,0 +1,520 @@
...
+      - The storage type/protocol to use when adding the storage.
+    type: str
+    required: true
+    choices: ['cephfs', 'cifs', 'iscsi', 'nfs', 'pbs']
+  cephfs_options:
+    description:
+      - Extended information for adding CephFS storage.
+    type: dict
+    suboptions:
+      monhost:
```
Thulium-Drake, line 52: This can be multiple in my cluster it looks like this in the WebUI, so it might be good to allow people to put in a list.

Also, there's an option to have Proxmox manage the Ceph storage, but currently the module doesn't allow to create it. It looks like this in the WebUI

image

I was unable to find any docs on what API call is needed to create a storage like that (and using the dev console in Firefox doesn't help :-/ ). BUT, from what I was able to test, if the PVE API gets a request for creating a CephFS _without_ `monhost` defined, it assumes it's using it's own Ceph cluster.

Here's the task:

```
      - name: Create cephfs
        community.proxmox.proxmox_storage:
          api_user: "{{ pve_user }}"
          api_password: "{{ pve_password }}"
          api_host: "{{ pve_server }}"
          name: 'testceph'
          node_name: 'm-v02-02'
          nodes:
            - 'm-v02-01'
            - 'm-v02-02'
            - 'm-v02-03'
          state: 'present'
          type: 'cephfs'          
          cephfs_options:   
            path: '/mnt/pve/testceph'
          content:          
            - 'iso'
            - 'backup'
```

plugins/modules/proxmox_storage.py
gyptazy: Yes and no. Since we have a multi-master setup, we can still execute the things in a different context. However, for the storage part which is on the datacenter level it isn't really required and we're also already able to filter and limit the addicted system(s) by the key `nodes`.

plugins/modules/proxmox_storage.py
```diff
@@ -0,0 +1,520 @@
...
+      subdir:
+        description:
+          - The subdir to be used within the CephFS.
+        type: str
+        required: false
+      client_keyring:
+        description:
+          - The client keyring to be used.
+        type: str
+        required: false
```
Thulium-Drake, line 83: However, it's good to know that this will not create a new Ceph pool to store the data in. It will rather use an existing pool (the `fs-name` parameter in the API), which is missing

```suggestion
        required: false
      fs_name:
        description:
          - The Ceph filesystem name
        type: str
        required: false
```

plugins/modules/proxmox_storage.py
```diff
@@ -0,0 +1,520 @@
...
+            'content': content
+        }
+
+        # Validate required parameters based on storage type
+        if storage_type == "cephfs":
+            cephfs_options = self.module.params.get(f'{storage_type}_options', {})
+            monhost = cephfs_options.get('monhost')
+            username = cephfs_options.get('username')
+            password = cephfs_options.get('password')
+            path = cephfs_options.get('path', '/')
```
Thulium-Drake, line 297:
```suggestion
            path = cephfs_options.get('path', '/')
            fs_name = cephfs_options.get('fs_name', None)
```

plugins/modules/proxmox_storage.py
```diff
@@ -0,0 +1,492 @@
...
+            client_keyring = cephfs_options.get('client_keyring')
+            if not all([monhost, path]):
+                self.module.fail_json(msg="CephFS storage requires 'monhost' and 'path' parameter.")
+            else:
+                payload['monhost'] = monhost
+                if username:
+                    payload['username'] = username
+                if password:
+                    payload['password'] = password
+                payload['path'] = path
```
Thulium-Drake:
```suggestion
                if fs_name:
                    payload['fs-name'] = fs_name
                payload['path'] = path
```

Thulium-Drake:
> > Say you only want to update the contents for an existing storage. How would what be handled (e.g. some storage named 'data' on which you want to enable ISO storage)?
> > I'd be OK with the answer being 'define everything', but it might be something to think about :-)
> 
> I also had this in mind, but unfortunately we have no patch methods within the Proxmox API for storage included. This means, that we get immediately an error returned, when the storage id (name) is already in use: `Failed to create storage: 500 Internal Server Error: create storage failed: storage ID 'net-nfsshare01' already defined"`.
> 
> As a result, we have the following possibilities:
> 
>     * Deal with it and keep it as it is. The ID is a unique constraint that can not be altered. Therefore, it must be intentionally removed and re-added when already present.
>
Nah, I just looked at the storage API docs, it is what it is, I think we should just do it like this -> Have the user define the truth and nothing but the truth ( ;-) ) when updating a storage config. This makes sense from a desired-state point of view, as this is the only way a storage will fully conform to the desired state.
> 
>     * We force remove the storage, update the values and add it within the module. It's not a big deal, but as an Ansible user I would not expect that updating a content parameter would remove me the whole storage from the cluster - which might even create a huge mess in production setups because of an unexpected event.
>
This might have some Unforeseen Consequences when dealing with network attached storages, some will be implemented on the host the moment they are changed/removed.
> 
>     * We could update the content within the storage configuration but it would omit the API. I'd highly like to only use the API methods to make sure we do not need any SSH access to systems using those modules.
> 
> 
> Happy to hear how you'd like to proceed.

plugins/modules/proxmox_storage.py
gyptazy: > This can be multiple in my cluster it looks like this in the WebUI, so it might be good to allow people to put in a list.

Good catch. According (sadly) to the API to be defined as a str.

gyptazy:
Great review @Thulium-Drake, thanks for your efforts and CephFS tests! Can you maybe test the Proxmox own CephFS integration by the last commit (also includes all other requested changes), please?

My assumption is, that we:
* Have a list of monhosts
* If no monhost we use an empty string
    * I'm not sure if it requires to not be delivered or an empty string (see also:; https://github.com/ansible-collections/community.proxmox/compare/88ef29cdc8bd0d12ada6b1b9153b3c3d732fb606..0dc4fc6d8011419d7984401d690fd05caad42369)
* If we have one or monhosts
    * We split them to strings according to the API documentation

plugins/modules/proxmox_storage.py
```diff
@@ -0,0 +1,502 @@
...
+
+class ProxmoxNodeAnsible(ProxmoxAnsible):
+    def add_storage(self):
+        changed = False
+        result = "Unchanged"
+        storage_name = self.module.params['name']
+        storage_type = self.module.params['type']
+        nodes = self.module.params['nodes']
+        content = self.module.params.get('content')
+        content = ",".join(content)
```
Thulium-Drake:
```suggestion

```

plugins/modules/proxmox_storage.py
```diff
@@ -0,0 +1,502 @@
...
+
+        # Create payload for storage creation
+        payload = {
+            'storage': storage_name,
+            'type': storage_type,
+            'nodes': ','.join(nodes),
```
Thulium-Drake:
```suggestion
            'nodes': nodes,
```

plugins/modules/proxmox_storage.py
```diff
@@ -0,0 +1,502 @@
...
+            monhost = cephfs_options.get('monhost', '')
+            username = cephfs_options.get('username')
+            password = cephfs_options.get('password')
+            path = cephfs_options.get('path', '/')
+            subdir = cephfs_options.get('subdir', None)
+            client_keyring = cephfs_options.get('client_keyring')
+            fs_name = cephfs_options.get('fs_name')
+
+            if not monhost == "":
+                payload['monhost'] = ",".join(monhost)
```
Thulium-Drake:
```suggestion
                payload['monhost'] = monhost
```

Thulium-Drake (changes requested):
I ran into some issues testing the module with the latest commit. But I removed all `join` statements, the API seems able to interpret the lists on it's own.

The next thing I'm looking into is having `nodes` and `content` only required when `state` is `present`

gyptazy:
> The next thing I'm looking into is having `nodes` and `content` only required when `state` is `present`

Adjusted to take only place when adding storage.

Thulium-Drake:
Awesome, this works really well, I've been using the playbook below:

```
---
- name: 'Create storage'
  hosts: 'localhost'
  connection: 'local'
  become: false
  gather_facts: false
  vars:
    pve_password: 'secret'
    pve_user: 'root@pam'
    pve_server: 'm-v02-01.rh.lab'
  tasks:
    - name: Create cephfs
      community.proxmox.proxmox_storage:
        api_user: "{{ pve_user }}"
        api_password: "{{ pve_password }}"
        api_host: "{{ pve_server }}"
        name: 'testceph'
        nodes:
          - 'm-v02-01'
          - 'm-v02-02'
          - 'm-v02-03'
        state: 'present'
        type: 'cephfs'
        cephfs_options:
          path: '/mnt/pve/testceph'
        content:
          - 'iso'
          - 'backup'

    - name: Wait a bit
      ansible.builtin.pause:
        seconds: 10

    - name: Delete cephfs
      community.proxmox.proxmox_storage:
        api_user: "{{ pve_user }}"
        api_password: "{{ pve_password }}"
        api_host: "{{ pve_server }}"
        name: 'testceph'
        type: 'cephfs'
        state: 'absent'
```

NOTE: I haven't tested this with an external Ceph, because I don't have one available :-)

With regards to the change, there's a way to make Ansible handle required args when needed with `required_if`. https://docs.ansible.com/ansible/latest/dev_guide/developing_program_flow_modules.html#dependencies-between-module-options

gyptazy:
Hey @Thulium-Drake,

> Awesome, this works really well, I've been using the playbook below:

Great, really happy to hear! Thank you for processing the Ceph based tests - appreciate this. And it turns out, it was needed - simply doing it in theory wouldn't fit ;) Thanks!

> With regards to the change, there's a way to make Ansible handle required args when needed with `required_if`. https://docs.ansible.com/ansible/latest/dev_guide/developing_program_flow_modules.html#dependencies-between-module-options


Tasks:
```
- name: Add NFS storage to Proxmox VE Cluster
  community.proxmox.proxmox_storage:
    api_host: 10.10.10.91
    api_user: root@pam
    api_password: admin123
    validate_certs: false
    nodes: ["cd-demo01"]
    state: present
    type: nfs
    name: net-nfsshare01
    nfs_options:
      server: 10.10.10.94
      export: "/mnt/storage01/s01nfs01"
    #content: ["rootdir", "images", "iso"]
  delegate_to: localhost
```

Without `content`:
```
TASK [dev : Add NFS storage to Proxmox VE Cluster] *************************************************************************************************************************************************************************************************
fatal: [cd-demo01.lab.gyptazy.com -> localhost]: FAILED! => {"changed": false, "msg": "state is present but all of the following are missing: content"}
```

With `content`:
```
TASK [dev : Add NFS storage to Proxmox VE Cluster] *************************************************************************************************************************************************************************************************
changed: [cd-demo01.lab.gyptazy.com -> localhost]
```

Thanks,
gyptazy

plugins/modules/proxmox_storage.py
```diff
@@ -0,0 +1,504 @@
...
+      portal: 10.10.10.94
+      target: "iqn.2005-10.org.freenas.ctl:s01-isci01"
+    content: ["rootdir", "images"]
+- name: Remove storage from Proxmox VE Cluster
+  community.proxmox.proxmox_storage:
+    api_host: proxmoxhost
+    api_user: root@pam
+    api_password: password123
+    validate_certs: false
+    nodes: ["de-cgn01-virt01", "de-cgn01-virt02"]
```
Thulium-Drake:
```suggestion
```

plugins/modules/proxmox_storage.py
```diff
@@ -0,0 +1,504 @@
...
+    state: absent
+    name: net-nfsshare01
+    type: nfs
+    server: 10.10.10.94
```
Thulium-Drake:
```suggestion
```

plugins/modules/proxmox_storage.py
```diff
@@ -0,0 +1,504 @@
...
+    content: ["rootdir", "images"]
```
Thulium-Drake:
```suggestion
```

Thulium-Drake:
With regards to the unit tests, which API call answers can I provide to test ceph?

Also I think we can handle adding the rest of the storage options later

gyptazy:
> With regards to the unit tests, which API call answers can I provide to test ceph?

I'm not sure if I got it correctly but I've just added a test for this. You can use the following params:
```
    cephfs_args = {
        "api_host": "localhost",
        "api_user": "root@pam",
        "api_password": "secret",
        "validate_certs": False,
        "node_name": "pve01",
        "nodes": ["pve01", "pve02"],
        "state": "present",
        "name": "cephfs-storage",
        "type": "cephfs",
        "cephfs_options": {
            "monhost": ["10.0.0.1", "10.0.0.2"],
            "username": "admin",
            "password": "secretpass",
            "path": "/",
            "subdir": "mydata",
            "client_keyring": "AQ==",
            "fs_name": "mycephfs"
        },
        "content": ["images", "rootdir"]
    }
```

I hope this was the intention and I did not misunderstood it.

Thulium-Drake:
It's OK, I wanted to know if you needed any API call replies when creating a cephfs in order to make it into a unit test :-) but I don't have an external ceph cluster. I did get the response from the API when creating an integrated CephFS

Request (using WebUI):
```
storage	"test"
fs-name	"data"
nodes	""
content	"backup"
type	"cephfs"
disable	"0"
prune-backups	"keep-all=1"
```

Response:
```
success	1
data	{ storage: "test", type: "cephfs" }
```

I also checked how it looked for other storages, it's the exact same. Which is rather sparse. So I think making unit tests for this API isn't as useful as you might think :zany_face: 

Perhaps the value for unit tests would be verifying the input it coherent with the samples created?

If not, I'm fine with merging this module as is, I have verified it's functionality by hand as well. And it's a very good candidate for integration testing :-)

gyptazy:
Hey @Thulium-Drake!

Thanks for your reply, I think we can go with it as it is after you also validated it by hand (I did it also for NFS, iSCSI & PBS).

Regarding integration tests, I need to have a further look at it - I must also admit that this is currently not my highest priority rather than getting all the "missing" modules to upstream to maintain them at a single place and habe them available to everyone. Would you mind to do it in another PR?

Cheers,
gyptazy

Thulium-Drake:
Integration tests are still something I'm working on with @kevinquick we'll be setting up a Proxmox server in the (near) future that will host these tests.

So yes, a new PR is more then fine for that.

I appreciate the work you've done so far! :-)

gyptazy:
> Integration tests are still something I'm working on with @kevinquick we'll be setting up a Proxmox server in the (near) future that will host these tests.

If you need resources for hosting a PVE node for integration tests (guess minimum would already fit), I can help you out at my BoxyBSD.com project (but only public IPv6). I’m in the Ansible dev Matrix channel or via https://gyptazy.com/#contacts-e 

> So yes, a new PR is more then fine for that.
> I appreciate the work you've done so far! :-)

Thanks, just ping me if I can assist :)

PR #153, opened 2025-08-01
Feature: add proxmox_node_network and proxmox_node_network_info modules

aleskxyz:
##### SUMMARY
Add comprehensive network interface management capabilities to the Proxmox collection with two new modules: `proxmox_node_network` for managing network interfaces and `proxmox_node_network_info` for retrieving network interface information. These modules support all major Proxmox network interface types including bridges, bonds, VLANs, and Open vSwitch interfaces with staged configuration management to prevent accidental network disruption.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
proxmox_node_network, proxmox_node_network_info

##### ADDITIONAL INFORMATION
**New Modules Added:**
- `proxmox_node_network`: Manage network interfaces (bridge, bond, eth, vlan, OVS)
- `proxmox_node_network_info`: Retrieve network interface information

**Key Features:**
- Staged configuration management with apply/revert functionality
- Support for all major interface types (bridge, bond, eth, vlan, OVSBridge, OVSBond, OVSIntPort)
- Comprehensive parameter validation for each interface type
- Support for parameter deletion (string params with empty string, integer params with -1)
- Full check mode support
- Full diff mode support

### DIFF
--- a/plugins/modules/proxmox_node_network.py
+++ b/plugins/modules/proxmox_node_network.py
@@ -1,0 +1,1769 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2025, aleskxyz <aleskxyz@gmail.com>
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+DOCUMENTATION = r"""
+module: proxmox_node_network
+version_added: "1.4.0"
+short_description: Manage network interfaces on Proxmox nodes
+requirements:
+  - ipaddress
+description:
+  - This module allows you to manage network interfaces on Proxmox nodes.
+  - Supports various interface types including bridge, bond, eth, vlan, and OVS interfaces.
+  - Network configuration changes are staged and must be explicitly applied using C(state=apply) to take effect.
+  - Changes made with C(state=present) or C(state=absent) are only staged and will not affect the running network configuration until applied.
+author: "aleskxyz (@aleskxyz)"
+notes:
+  - Network configuration changes are staged and must be explicitly applied using C(state=apply) to take effect.
+  - Changes made with C(state=present) or C(state=absent) are only staged and will not affect the running network configuration until applied.
+  - Interface type cannot be changed after creation. Delete and recreate the interface to change its type.
+  - "Ethernet interfaces (C(eth) type) represent physical hardware and cannot be created via API. Only existing physical interfaces can be configured."
+  - "Bond and OVS bond interfaces must follow naming format C(bondX) where X is a number between 0 and 9999 (e.g., C(bond0), C(bond1), C(bond9999))."
+  - "VLAN interfaces support two naming formats: C(vlanXY) (e.g., C(vlan100)) and C(iface_name.vlan_id) (e.g., C(eth0.100))."
+  - For C(vlanXY) format, C(vlan_raw_device) parameter is required.
+  - For C(iface_name.vlan_id) format, C(vlan_raw_device) should not be specified.
+  - "Parameter deletion is supported for specific parameters. Set string parameters to empty string C('') or integer parameters to C(-1) to delete them."
+  - "Deletable parameters: C(cidr), C(gateway), C(cidr6), C(gateway6), C(comments), C(mtu),
+    C(bridge_ports), C(bridge_vids), C(ovs_ports), C(ovs_options), C(ovs_tag)."
+  - "When C(state=apply) or C(state=revert), only the C(node) parameter is accepted. All other parameters are not allowed."
+seealso:
+  - module: community.proxmox.proxmox_node_network_info
+    description: Retrieve information about network interfaces on Proxmox nodes.
+  - name: Proxmox Network Configuration
+    description: Proxmox VE network configuration documentation.
+    link: https://pve.proxmox.com/wiki/Network_Configuration
+attributes:
+  check_mode:
+    support: full
+    details: Check mode is fully supported for all states.
+  diff_mode:
+    support: full
+    details: Check mode is fully supported for all states.
+options:
+  node:
+    description:
+      - The Proxmox node to manage network interfaces on.
+    type: str
+    required: true
+  state:
+    description:
+      - The desired state of the network interface.
+      - C(present) and C(absent) stage changes but do not apply them to the running configuration.
+      - C(apply) applies all staged network configuration changes to the running configuration.
+      - C(revert) discards all staged network configuration changes.
+    type: str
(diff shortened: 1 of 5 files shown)


Thulium-Drake:
@gyptazy Would you have time to review this? I'm not well-versed enough in Python to review this :-(

gyptazy:
> @gyptazy Would you have time to review this? I'm not well-versed enough in Python to review this :-(

Will do but can't promise to do it this weekend. Probably on Monday.

plugins/modules/proxmox_node_network.py
```diff
@@ -0,0 +1,1734 @@
...
+        "read_only": True,
+    },
+}
+
+
+def _is_valid_cidr(cidr):
+    """Validate IPv4 CIDR format."""
+    if not cidr:
+        return False
+    pattern = r"^(\d{1,3}\.){3}\d{1,3}/\d{1,2}$"
```
gyptazy: Kudos for validating! But in that case we should probably to a more strict validating like:
```
_cidr_re = re.compile(
    r'^(25[0-5]|2[0-4]\d|1\d{2}|[1-9]?\d)\.'
    r'(25[0-5]|2[0-4]\d|1\d{2}|[1-9]?\d)\.'
    r'(25[0-5]|2[0-4]\d|1\d{2}|[1-9]?\d)\.'
    r'(25[0-5]|2[0-4]\d|1\d{2}|[1-9]?\d)'
    r'/(3[0-2]|[12]?\d)$'
)
```

Maybe this is the point, where we should simply use libs like `ipaddress` instead of reinventing the wheel.

plugins/modules/proxmox_node_network.py
```diff
@@ -0,0 +1,1734 @@
...
+    return bool(re.match(pattern, cidr))
+
+
+def _is_valid_cidr6(cidr):
+    """Validate IPv6 CIDR format."""
+    if not cidr:
+        return False
+    pattern = r"^[0-9a-fA-F:]+/\d{1,3}$"
```
gyptazy: That might not be strict enough. Maybe this fits better:
```
r'^([0-9A-Fa-f:]+::?[0-9A-Fa-f:]*)/(12[0-8]|1[01][0-9]|[0-9]{1,2})$'
```
Also here, maybe better to use an external lib like `ipaddress`; even I'd like to avoid it.

gyptazy:
Hey @aleskxyz,

that's well written and structured - appreciate your work! There're just some smaller things that're not yet important (but might need some changes in future releases) but it doesn't make sense or brings any huge benefits to adjust this right now. I didn't give OVS a try but apart from that it looks good to me. Just the regex might fail. I mean, it would be a user error but when we already have a validate function, we should do a proper validation (I added a quick idea, could probably also be improved).

So far, well done!

cc: @Thulium-Drake 

Cheers,
gyptazy

plugins/modules/proxmox_node_network.py
IamLunchbox: And since `ipaddress` is part of the standard library, using it would outsource complexity and not include major (installation) overhead. :)

Out of curiosity: Why would you like avoid including this library otherwise?

aleskxyz:
> Hey @aleskxyz,
> 
> that's well written and structured - appreciate your work! There're just some smaller things that're not yet important (but might need some changes in future releases) but it doesn't make sense or brings any huge benefits to adjust this right now. I didn't give OVS a try but apart from that it looks good to me. Just the regex might fail. I mean, it would be a user error but when we already have a validate function, we should do a proper validation (I added a quick idea, could probably also be improved).
> 
> So far, well done!
> 
> cc: @Thulium-Drake
> 
> Cheers, gyptazy

Hi @gyptazy 
I have updated the code and refactored the logic with the help of the ipaddress module. I have also mentioned it as a requirement in the document section.
The feature branch has been rebased onto the latest commit on the main branch.
Thanks.

Thulium-Drake:
@aleskxyz Thanks for your work! 

@gyptazy @IamLunchbox Are you happy with the results? :-)

gyptazy:
> @aleskxyz Thanks for your work!

Also from me, thanks again for your work and sorry for the delay!

> @gyptazy @IamLunchbox Are you happy with the results? :-)

I’m currently abroad and will have a look at it asap.

plugins/modules/proxmox_node_network.py
```diff
@@ -0,0 +1,1769 @@
...
+    choices:
+      - layer2
+      - layer2+3
+      - layer3+4
+    required: false
+  slaves:
+    description:
+      - Interfaces used by the bonding device (space separated).
+      - Required for C(bond) interface type.
+    type: str
```
gyptazy, line 202: We should probably move to a list object here (instead of comma separated lists of multiple slave interfaces).

gyptazy:
Hey everyone,

thanks for adjusting my remarks @aleskxyz and I think this makes it much simpler to use, read and evaluate when running into issues. For me, it looks good so far from reading the code (since I'm still abroad I don't have any access to test the code), so I couldn't give it a try, yet.

Just a small note - there're some "code redundancies" like:
```
def _is_valid_cidr(cidr):
    """Validate IPv4 host address with prefix using ipaddress."""
    if not cidr:
        return False
    try:
        iface = ip_interface(cidr)
        return iface.version == 4
    except Exception:
        return False


def _is_valid_cidr6(cidr):
    """Validate IPv6 host address with prefix using ipaddress."""
    if not cidr:
        return False
    try:
        iface = ip_interface(cidr)
        return iface.version == 6
    except Exception:
        return False
```

This could be simplified like the following (by setting the func arg from external functions when required):
```
def is_valid_cidr(cidr, cidr_type=4):
    """Validate IPv4 or IPv6 host address with prefix using ipaddress."""
    if not cidr:
        return False
    try:
        iface = ip_interface(cidr)
        return iface.version == cidr_type
    except ValueError:
        return False
```

But this isn't a showstopper for me.

Thulium-Drake:
@gyptazy how would you like to proceed? Are the modules fit for use (so we can process the changes you requested in later PRs if needed?)?

gyptazy:
Good morning @Thulium-Drake, hey everyone,

> @gyptazy how would you like to proceed? Are the modules fit for use (so we can process the changes you requested in later PRs if needed?)?

Yes, I think we can go this way. It’s a new module, so it wouldn’t break anything and imho we can safely merge it. I’d be happy to see it merged (as said, the things aren’t any show stopper). It can be improved afterwards. The new networking and firewall modules provide a great addition to the collection.

Thanks to everyone creating, contributing and reviewing the things to make our life’s easier :)

cheers,
gyptazy

Thulium-Drake:
Done! @aleskxyz, @gyptazy thanks for all the work you've done! <3

aleskxyz:
Thank you, @Thulium-Drake, for merging the PR.
It seems the new version is released, but this module is not included😕
Is there any issue with it?

IamLunchbox:
Yes, the changelog is empty as well. Something in the release pipeline failed?

@Thulium-Drake bump :)

PR #155, opened 2025-08-08
Fix AttributeError in ProxmoxAnsible initialization

kevinquick:
##### SUMMARY
  The module attribute was referenced before assignment when checking
  proxmoxer version, causing initialization to fail. Changed to use
  the module parameter directly before self.module is set.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
module_utils/proxmox.py


Thulium-Drake:
Thanks for your work! :grin:

PR #162, opened 2025-08-20
Allow import content templates

### LINKED ISSUE #161: Content Type 'import' not allowed by proxmox_template
##### SUMMARY

The proxmox API has endpoints for adding/deleting storage content here https://pve.proxmox.com/pve-docs/api-viewer/index.html#/nodes/{node}/storage/{storage}

Right now, community.proxmox only has a module to manage the storages, but not their content.

I'd like to have a module that allows me to upload ISO files or disk base images.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

`community.proxmox.proxmox_storage_contents` I guess? There already is an info module with that name so that would be consistent. I don't like the plural here though

##### ADDITIONAL INFORMATION

```yaml
- name: Upload local ISO to Proxmox VE Cluster storage, even if ISO with same name and size present
  community.proxmox.proxmox_storage_contents:
    api_host: proxmoxhost
    api_user: root@pam
    api_password: password123
    validate_certs: false
    node: dev01
    storage: local
    state: present
    filename: proxmox-install.iso
    content: iso
    file: "localpath/proxmox.iso"
    checksum_algorithm: sha256
    checksum: "042aea10a0f14f2d391373599be69d53a75dde9951fc3d3cd10b6100aa7a9f24"
    force: true

- name: Download Importable Disk to Proxmox VE Cluster storage
  community.proxmox.proxmox_storage_contents:
    api_host: proxmoxhost
    api_user: root@pam
    api_password: password123
    validate_certs: false
    node: dev01
    storage: local
    state: present
    filename: debian13.qcow2
    content: import
    url: "https://cloud.debian.org/images/cloud/trixie/latest/debian-13-genericcloud-arm64.qcow2"
    url_verify_certs: false
    checksum_algorithm: sha256
    checksum: "042aea10a0f14f2d391373599be69d53a75dde9951fc3d3cd10b6100aa7a9f24"
```

koalajoe23:
##### SUMMARY

Fixes #161 

Add 'import' to allowed content types of proxmox_template, so disk images and can be used as disk images on VM creation.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

proxmox_template

##### ADDITIONAL INFORMATION

- Simple one-line fix as they behave like ISOs
- Added examples as on a glance the module looks like only supporting LXC templates

### DIFF
--- a/plugins/modules/proxmox_template.py
+++ b/plugins/modules/proxmox_template.py
@@ -168,6 +168,24 @@ EXAMPLES = r"""
     url: ubuntu-20.04-standard_20.04-1_amd64.tar.gz
     checksum_algorithm: sha256
     checksum: 65d860160bdc9b98abf72407e14ca40b609417de7939897d3b58d55787aaef69
+
+- name: Upload new disk image template with minimal options
+  community.proxmox.proxmox_template:
+    node: uk-mc02
+    api_user: root@pam
+    api_password: 1q2w3e
+    api_host: node1
+    content_type: import
+    src: debian-13-genericcloud-amd64.qcow2
+
+- name: Upload new ISO with minimal options
+  community.proxmox.proxmox_template:
+    node: uk-mc02
+    api_user: root@pam
+    api_password: 1q2w3e
+    api_host: node1
+    content_type: import
+    src: proxmox.iso
 """
 
 import os
--- a/plugins/modules/proxmox_template.py
+++ b/plugins/modules/proxmox_template.py
@@ -184,7 +184,7 @@ EXAMPLES = r"""
     api_user: root@pam
     api_password: 1q2w3e
     api_host: node1
-    content_type: import
+    content_type: iso
     src: proxmox.iso
 """
 
--- a/plugins/modules/proxmox_template.py
+++ b/plugins/modules/proxmox_template.py
@@ -42,11 +42,11 @@ options:
     type: str
   content_type:
     description:
-      - Content type.
-      - Required only for O(state=present).
+      - Content type of the template.
+      - Valid values are V(vztmpl) for LXC container templates, V(iso) for ISO disc images, and V(import) for harddisk images and Open Virtual Appliances (OVA).
     type: str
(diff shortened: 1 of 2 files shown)


plugins/modules/proxmox_template.py
```diff
@@ -168,6 +168,24 @@
...
+    content_type: import
+    src: debian-13-genericcloud-amd64.qcow2
+
+- name: Upload new ISO with minimal options
+  community.proxmox.proxmox_template:
+    node: uk-mc02
+    api_user: root@pam
+    api_password: 1q2w3e
+    api_host: node1
+    content_type: import
```
Thulium-Drake:
```suggestion
    content_type: iso
```

plugins/modules/proxmox_template.py
```diff
@@ -42,11 +42,11 @@
     type: str
   content_type:
     description:
-      - Content type.
-      - Required only for O(state=present).
```
Thulium-Drake, line 46: Does this mean that removing content also requires a `content_type` now? If so, does the module handle 'legacy' playbooks gracefully, or is this a breaking change?

Thulium-Drake:
Also, can you add a changelog fragment?

plugins/modules/proxmox_template.py
koalajoe23: Sorry for leaving no explanation, had to retrace my thought process too :) I deleted line 46 because it is not correct.

If I don't miss anything here and according to my tests, `content_type` is neither explicitly needed for present and absent `state`. Both times `content_type` will fallback to the `vztmpl` value.

While I'd rather have no default type for this and to have the user  declare it explicitly, I understand if that is not feasible due to *this* being a breaking change.

koalajoe23:
> Also, can you add a changelog fragment?

Done, hopefully as intended, pls correct if not.

Thulium-Drake:
Thanks for this! :grin:

PR #167, opened 2025-08-20
feature: add new module proxmox_cluster_ha_rules

rekup:
##### SUMMARY
This pull request adds a new module to support ha rules which were introduced in PVE 9

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME

proxmox_cluster_ha_rules

##### ADDITIONAL INFORMATION

New Modules Added:

* proxmox_cluster_ha_rules: Manage HA rules in PVE9 or higher

The new feature is described in the [High Availability Guide](https://pve.proxmox.com/wiki/High_Availability) as well as the [API documentation](https://pve.proxmox.com/pve-docs/api-viewer/#/cluster/ha/rules/{rule}).

I was not sure how to implement the parameters, which have a default value on the api itself. For example, the parameter `strict` has a default of `0`. If we set a default of `strict=false` in the module, a rule which is currently defined with `strict=1` will change to `strict=0` if a user removes the parameter in the ansible task. I therefore chose to not set a default in the ansible module and let the api handle this.

Another thing I was not sure about was the implementation of the `affinity`, `nodes` and `resources`. These are basically required for every new rule. But if we only want to update the `comment` of an existing rule, a user might not expect them as required. I see basically two ways to implement this:

1. Always mark the parameters as required
2. Implement some logic and use `fail_json` if a user tries to create a rule (which does not exist yet) without these parameters.

I chose to go with option 2, but I'm open for other solutions

### DIFF
--- a/meta/runtime.yml
+++ b/meta/runtime.yml
@@ -14,6 +14,7 @@ action_groups:
     - proxmox_backup_schedule
     - proxmox_cluster
     - proxmox_cluster_ha_resources
+    - proxmox_cluster_ha_rules
     - proxmox_cluster_ha_groups
     - proxmox_cluster_join_info
     - proxmox_disk
--- a/plugins/modules/proxmox_cluster_ha_rules.py
+++ b/plugins/modules/proxmox_cluster_ha_rules.py
@@ -82,7 +82,7 @@ options:
         type: bool
     type:
         description: HA rule type.
-        required: true
+        required: false
         choices: ['node-affinity', 'resource-affinity']
         type: str
 
@@ -161,18 +161,6 @@ EXAMPLES = r"""
       - vm:100
       - vm:101
     disable: false
-  delegate_to: localhost
-
-- name: Update the comment of an existing rule - this will fail if the rule does not exist yet
-  community.proxmox.proxmox_cluster_ha_rules:
-    api_host: "{{ proxmox_api_host }}"
-    api_user: "{{ proxmox_api_user }}"
-    api_token_id: "{{ proxmox_api_token_id }}"
-    api_token_secret: "{{ proxmox_api_token_secret }}"
-    name: resource-affinity-rule-1
-    state: present
-    type: resource-affinity
-    comment: My new description
   delegate_to: localhost
 """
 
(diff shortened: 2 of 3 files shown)


Thulium-Drake:
Hi there :wave: ,

Thanks for this!  With regards to defaults, please let the API handle that, just make sure that the module handles any failures with grace (and tell the user what's wrong) :-)

And with regards to making the `affinity`, `nodes` and `resources` required, if the API expects them, we should send them. IMHO, you shouldn't make tasks that only update the `comment`, especially due to the idempotent nature of Ansible. I'd rather have the module handles administering rules in an idempotent manner (having diff support is a nice extra here).

Cheers!

rekup:
Hi :wave: 

Sounds good to me, I updated the module and made the `type` and `resources` params required if `state=present` (only the `name` is used to uniquely identify a rule so making these params required if `state=absent` does not do anything).

Regarding the defaults: I agree that this should be handled by the api. But this gets a bit tricky with the `disable` param. If a rule is enabled (-> `disable=false`) the api response is missing this parameter completely. This poses a slight problem because if a user creates a rule with `disable=false` and the rule is already enabled (api response is missing the `disable` parameter) we somehow need to handle that in the module otherwise this will break idempotency:

```
TASK [proxmox : Enable a rule which is already enabled] *****************************************************************************************************************************************************************
--- before
+++ after
@@ -1,6 +1,7 @@
 {
     "comment": "rku test 1",
     "digest": "f8ec0f407d3cbfe8db2ab5c509499bceedf7f7f7",
+    "disable": 0,
     "nodes": "proxmox01:10,proxmox02:20",
     "order": 2,
     "resources": "vm:102",
```

The only way I can think of (without breaking idempotency) is to set a default using something like `existing_rule.setdefault("disable", 0)`. Without that, a task (with `disable=false`) which updates an already enabled a rule will show up as `changed` every time. A user could then remove the `disable` parameter from the task to "fix" this but that seems a bit unintuitive. The same also applies to the `comment` parameter

Thulium-Drake:
That sounds sensible to me, because it sounds to me like the API _will_ respond with `disable=true` when it is in fact disabled. Therefore I think it's OK to assume that and make the code substitute the missing variable.

And if the Proxmox devs decide to add this field, just don't replace it ;-)

Thulium-Drake:
@rekup LGTM, is there anything else you'd like to change/add to the rule, or can I merge it?

rekup:
Thanks for the review! In my opinion, this PR can be merged

PR #172, opened 2025-08-26
Added link0 and link1 to join command, also reworked building payload…

### LINKED ISSUE #168: Separated cluster network not working with cluster join
Hello,

When doing a cluster_join, if we want to use a separate cluster network, the link0 parameter is not used so the "public" ip address is used.
When creating the cluster, the link0 parameter is perfectly used, so for the "master", the separated cluster network is correctly used.
in cluster_create function, we can see this (which is not seen in cluster_join function) : 
if self.module.params.get("link0") is not None:
            payload["link0"] = self.module.params.get("link0")

Is it possible to add this "feature" ?
thanks !

Thulium-Drake:
##### SUMMARY

Added link0 and link1 to join command, also reworked building payload to be consistent with creating clusters
Fixes #168 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

proxmox_cluster
##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/168-add-link-arguments-to-cluster.yaml
+++ b/changelogs/fragments/168-add-link-arguments-to-cluster.yaml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - proxmox_cluster - added link0 and link1 to join command (https://github.com/ansible-collections/community.proxmox/issues/168, https://github.com/ansible-collections/community.proxmox/pull/172).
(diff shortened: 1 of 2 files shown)


Thulium-Drake:
@gyptazy thanks for the good unit tests :sunglasses:

gyptazy:
looks much cleaner with the payload approach :)

Thanks for doing so!

PR #174, opened 2025-08-28
Add ignores for ansible.module_utils.six

felixfontein:
##### SUMMARY
Ref: https://github.com/ansible/ansible/pull/85651

Since this collection no longer supports Python 2.7 it would be better to get rid of these imports instead, but as a first step this fixes CI (which is currently broken due to the ansible-core PR).

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
sanity test ignore files


felixfontein:
Closing in favor of #175.

PR #176, opened 2025-08-30
proxmox_zone - new module

JanaHoch:
##### SUMMARY

New module to list/creat/update/delete SDN zones

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

proxmox_zone

### DIFF
--- a/plugins/modules/proxmox_zone.py
+++ b/plugins/modules/proxmox_zone.py
@@ -9,6 +9,9 @@ from __future__ import absolute_import, division, print_function
 
 __metaclass__ = type
 
+# from ansible_collections.community.sap_libs.plugins.modules.sap_control_exec import choices
+# from pygments.lexer import default
+
 DOCUMENTATION = r""""""
 
 EXAMPLES = r""""""
@@ -21,46 +24,159 @@ from ansible_collections.community.proxmox.plugins.module_utils.proxmox import (
     ProxmoxAnsible
 )
 
+def get_proxmox_args():
+    return dict(
+        state=dict(type="str", choices=["present", "absent"], required=False),
+        force=dict(type="bool", default=False, required=False),
+        update=dict(type="bool", default=False, required=False),
+        type=dict(type="str",
+                  choices=["evpn", "faucet", "qinq", "simple", "vlan", "vxlan"],
+                  required=False),
+        zone=dict(type="str", required=False),
+        advertise_subnets=dict(type="bool", required=False),
+        bridge=dict(type="str", required=False),
+        bridge_disable_mac_learning=dict(type="bool", required=False),
+        controller=dict(type="str", required=False),
+        dhcp=dict(type="str", choices=["dnsmasq"], required=False),
+        disable_arp_nd_suppression=dict(type="bool", required=False),
+        dns=dict(type="str", required=False),
+        dnszone=dict(type="str", required=False),
+        dp_id=dict(type="int", required=False),
+        exitnodes=dict(type="str", required=False),
+        exitnodes_local_routing=dict(type="bool", required=False),
+        exitnodes_primary=dict(type="str", required=False),
+        fabric=dict(type="str", required=False),
+        ipam=dict(type="str", required=False),
+        lock_token=dict(type="str", required=False),
+        mac=dict(type="str", required=False),
+        mtu=dict(type="int", required=False),
+        nodes=dict(type="str", required=False),
+        peers=dict(type="str", required=False),
+        reversedns=dict(type="str", required=False),
+        rt_import=dict(type="str", required=False),
+        tag=dict(type="int", required=False),
+        vlan_protocol=dict(type="str", choices=["802.1q", "802.1ad"], required=False),
+        vrf_vxlan=dict(type="int", required=False),
+        vxlan_port=dict(type="int", required=False),
+    )
 
-class ProxmoxZoneAnsible(ProxmoxAnsible):  
-   def get_zones(self, type):
+def get_ansible_module():
+    module_args = proxmox_auth_argument_spec()
+    module_args.update(get_proxmox_args())
+
+    return AnsibleModule(
+        argument_spec=module_args,
+        required_if=[
+            ('state', 'present', ['type', 'zone']),
+            ('update', True, ['zone'])
+        ]
+    )
+
+class ProxmoxZoneAnsible(ProxmoxAnsible):
+    def __init__(self, module):
+        super(ProxmoxZoneAnsible, self).__init__(module)
+        self.params = module.params
+
+    def run(self):
+        state = self.params.get("state")
+        force = self.params.get("force")
+
+        if state == "present":
+            params = {
+                "type": self.params.get("type"),
+                "zone": self.params.get("zone"),
+                "advertise-subnets": self.params.get("advertise_subnets"),
+                "bridge": self.params.get("bridge"),
+                "bridge-disable-mac-learning": self.params.get("bridge_disable_mac_learning"),
+                "controller": self.params.get("controller"),
+                "dhcp": self.params.get("dhcp"),
+                "disable-arp-nd-suppression": self.params.get("disable_arp_nd_suppression"),
+                "dns": self.params.get("dns"),
+                "dnszone": self.params.get("dnszone"),
+                "dp-id": self.params.get("dp_id"),
+                "exitnodes": self.params.get("exitnodes"),
+                "exitnodes-local-routing": self.params.get("exitnodes_local_routing"),
+                "exitnodes-primary": self.params.get("exitnodes_primary"),
+                "fabric": self.params.get("fabric"),
+                "ipam": self.params.get("ipam"),
+                "lock-token": self.params.get("lock_token"),
+                "mac": self.params.get("mac"),
+                "mtu": self.params.get("mtu"),
+                "nodes": self.params.get("nodes"),
+                "peers": self.params.get("peers"),
+                "reversedns": self.params.get("reversedns"),
+                "rt-import": self.params.get("rt_import"),
+                "tag": self.params.get("tag"),
+                "vlan-protocol": self.params.get("vlan_protocol"),
+                "vrf-vxlan": self.params.get("vrf_vxlan"),
+                "vxlan-port": self.params.get("vxlan_port"),
+            }
+            self.zone_present(force, **params)
+
+        elif state == "update":
+            self.zone_update(
+
+            )
+        elif state == "absent":
+            self.zone_absent(
+
+            )
+        else:
+            zones = self.get_zones({'type': self.params.get('type')})
+            self.module.exit_json(
+                changed=False, msg=zones
+            )
+
+
+    def get_zones(self, type=dict()):
+        print("reached")
         try:
-            if type == "all":
-              zones = self.proxmox_api.cluster().sdn().zones().get()
-            else:
-              zones = self.proxmox_api.cluster().sdn().zones().get(type=type)
-            return zones
-                
+            return self.proxmox_api.cluster().sdn().zones().get(**type)
         except Exception as e:
             self.module.fail_json(
                 msg=f'Failed to retrieve zone information from cluster: {e}'
             )
 
+    def zone_present(self, force, **kwargs):
+        available_zones = {x["zone"]: x["type"] for x in self.get_zones()}
+        zone = kwargs.get("zone")
+        type = kwargs.get("type")
+
+        # Check if zone already exists
+        if zone in available_zones.keys() and force:
+            if type != available_zones[zone]:
+                self.module.fail_json(
+                    zone=zone,
+                    msg=f'zone {zone} exists with different type and we cannot change type post fact.'
+                )
+            else:
+                del kwargs['type']
+                self.zone_update(kwargs)
+        elif zone in available_zones.keys() and not force:
+            self.module.exit_json(
+                changed=False, zone=zone, msg=f'Zone {zone} already exists and force is false!'
+            )
+        else:
+            self.proxmox_api.cluster().sdn().zones().post(**kwargs)
+            self.module.exit_json(
+                changed=True, zone=zone, msg=f'Created new Zone - {zone}'
+            )
+
+    def zone_update(self):
+        pass
+
+    def zone_absent(self):
+        pass
+
 
 def main():
-    
-    module_args = proxmox_auth_argument_spec()
-    zone_args = dict(
-        type=dict(type="str", 
-                  choices=["evpn", "faucet", "qinq", "simple", "vlan", "vxlan", "all"], 
-                  default="all", required=False)
-    )
-    module_args.update(zone_args)
-
-    module = AnsibleModule(
-        argument_spec=module_args,
-        required_together=[("api_token_id", "api_token_secret")],
-        required_one_of=[("api_password", "api_token_id")],
-        supports_check_mode=True,
-    )
-
+    module = get_ansible_module()
     proxmox = ProxmoxZoneAnsible(module)
-    type = module.params['type']
-    results = {}
-    zones = proxmox.get_zones(type)
 
-    results['zones'] = zones
-    module.exit_json(**results)
+    try:
+        proxmox.run()
+    except Exception as e:
+        module.fail_json(msg=f'An error occurred: {e}')
 
 if __name__ == "__main__":
     main()
--- a/plugins/modules/proxmox_zone.py
+++ b/plugins/modules/proxmox_zone.py
@@ -111,13 +111,15 @@ class ProxmoxZoneAnsible(ProxmoxAnsible):
             "vxlan-port": self.params.get("vxlan_port"),
         }
 
+        if zone_params['lock-token'] is None and state is not None:
+            zone_params['lock-token'] = self.get_global_sdn_lock()
+
         if state == "present":
             self.zone_present(force, **zone_params)
 
         elif state == "update":
-            self.zone_update(**zone_params
+            self.zone_update(**zone_params)
 
-            )
         elif state == "absent":
             self.zone_absent(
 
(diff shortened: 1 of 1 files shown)


Thulium-Drake:
Thanks for this work! :-) Can you have a look at the sanity checks (for this and the other PRs you've created)

JanaHoch:
> Thanks for this work! :-) Can you have a look at the sanity checks (for this and the other PRs you've created)

I've fixed sanity issues in all PRs please check. let me know if anything else is needed. Thanks.

JanaHoch:
sorry 1 more question I was trying merge on my local repo and it is giving merge conflicts for `meta/runtime.yml` do we need to add all new modules to that or if that is not required I can remove it so you won't get those conflicts. 

because sanity check is giving this without that ` meta/runtime.yml: module 'proxmox_zone' is not part of 'proxmox' action group` but it's in white so not sure if it's really needed or not.

Thulium-Drake:
I'm not too familiar _why_, but what I know is that each PR that adds a new module should add it to that list in the `runtime.yml`. But if you're trying to merge it locally, you should do the following to avoid merge conflicts:

- Merge a branch into your `main` branch
- Checkout the second branch you want to merge
- Rebase it on the latest state of `main`
- Repeat

The reason you get merge conflicts is because in each branch you add one line _after_ the line that is in the current main. Which, after merging one of those branches, has changed into something else. So from git's perspective your branches want to change the same line.

JanaHoch:
> I'm not too familiar _why_, but what I know is that each PR that adds a new module should add it to that list in the `runtime.yml`. But if you're trying to merge it locally, you should do the following to avoid merge conflicts:
> 
>     * Merge a branch into your `main` branch
> 
>     * Checkout the second branch you want to merge
> 
>     * Rebase it on the latest state of `main`
> 
>     * Repeat
> 
> 
> The reason you get merge conflicts is because in each branch you add one line _after_ the line that is in the current main. Which, after merging one of those branches, has changed into something else. So from git's perspective your branches want to change the same line.

Ok. let me add it. and also let me check what I can do for unit tests.

JanaHoch:
Hi this is also ready. once this gets merged I'll rebase my vnet and subnet branches so they'll also get locking function for testing. Thanks.

Thulium-Drake:
Let's focus on this and #183 first then. I'm not in a hurry to get a release out. I do think after your PRs have been merged we have a nice juicy release! ;-)

Thulium-Drake:
@IamLunchbox Can you take a look at this again, if you're happy, I'm happy.

I don't use SDN/PVE firewalls (nor understand how they work :sweat_smile: ).

@JanaHoch Thanks for making the unit tests, they seem to pass just fine and look extensive!

JanaHoch:
Hi @Thulium-Drake / @IamLunchbox , Sorry I just saw that I forgot to convert few ansible boolean to proxmox bool. I've fixed it now with 7b247507ac18bd4231bff6702d74ef98f4ae439c . Let me know if  you see anything else I've missed. I used `ansible_to_proxmox_bool()` so it should still return `None` if values are not passed and it shouldn't cause any issue for zone type where those aren't supposed to be present.

IamLunchbox:
Just FYI:
I will need some time to review the PR's, at least until sunday.

JanaHoch:
> Just FYI: I will need some time to review the PR's, at least until sunday.

No worries. take your time. I'll also try to finish writing unit tests for remaining PRs. Thanks.

plugins/module_utils/proxmox.py
```diff
@@ -245,3 +245,67 @@ def get_storage_content(self, node, storage, content=None, vmid=None):
...
+    def get_global_sdn_lock(self):
+        """Acquire global SDN lock. Needed for any changes under SDN.
+
+        @return: lock-token
+        """
+        try:
+            return self.proxmox_api.cluster().sdn().lock().post()
+        except Exception as e:
+            self.module.fail_json(
+                msg=f'Failed to acquire global sdn lock {e}'
```
IamLunchbox: @Thulium-Drake The last time i contributed to community.general f-strings were not allowed due to backwards compatibility. But sanity-tests pass on this code. Do you know why? Has python2 compatibility been deprecated for this collection?

plugins/modules/proxmox_zone.py
```diff
@@ -0,0 +1,536 @@
...
+# Copyright (c) 2025, Jana Hoch <janahoch91@proton.me>
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+# from ansible_collections.community.sap_libs.plugins.modules.sap_control_exec import choices
+# from pygments.lexer import default
```
IamLunchbox: Delete?

plugins/modules/proxmox_zone.py
```diff
@@ -0,0 +1,536 @@
...
+    validate_certs: no
+
+- name: Get all simple zones
+  community.proxmox.proxmox_zone:
+    api_user: "root@pam"
+    api_password: "{{ vault.proxmox.root_password }}"
+    api_host: "{{ pc.proxmox.api_host }}"
+    validate_certs: no
+    type: simple
+  register: zones
```
IamLunchbox: This module tries to retreive information and change SDNs in one go. IMHO, this is counterintuitive and not granular. 

Since other modules within ansible core and in this collection are also splitting gathering information and changing things into two different modules, I would suggest doing the same here and creating module `zone_info`. As an example: The usage between gathering info and changing stuff is split between `proxmox_vm_info` and `proxmox_vm` or `slurp` and `copy`.

plugins/modules/proxmox_zone.py
```diff
@@ -0,0 +1,536 @@
...
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+# from ansible_collections.community.sap_libs.plugins.modules.sap_control_exec import choices
+# from pygments.lexer import default
+
+DOCUMENTATION = r"""
```
IamLunchbox, line 12: Semantics are a bit uneven in the documentation. Some descriptions start with lower case, some with upper. Some are terminated by a dot, others are not.

See also [this](https://docs.ansible.com/ansible/devel/dev_guide/style_guide/grammar_punctuation.html) guideline.

plugins/modules/proxmox_zone.py
```diff
@@ -0,0 +1,536 @@
...
+    description:
+      - The desired state of the zone configuration.
+    type: str
+    choices:
+      - present
+      - absent
+      - update
+  force:
+    description:
+      - If state is present and zone exists it'll update.
```
IamLunchbox: Use ansible markup for better representation in the ansible docs :)

See [here](https://docs.ansible.com/ansible/devel/dev_guide/ansible_markup.html#ansible-markup).

plugins/modules/proxmox_zone.py
```diff
@@ -0,0 +1,536 @@
...
+    description:
+      - Vxlan tunnel udp port (default 4789).
+    type: int
+extends_documentation_fragment:
+  - community.proxmox.proxmox.actiongroup_proxmox
+  - community.proxmox.proxmox.documentation
+  - community.proxmox.attributes
+"""
+
+EXAMPLES = r"""
```
IamLunchbox, line 159: @Thulium-Drake Should example tasks be added to the intergration tests only after #7 is solved?

IamLunchbox:
Meh, sorry for the spam! I should've done the reviews in one review-context I guess? 

Anyways, I am about 50% done now, mostly the documentation. Please remember to prepare a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment).

JanaHoch:
> Meh, sorry for the spam! I should've done the reviews in one review-context I guess?
> 
> Anyways, I am about 50% done now, mostly the documentation. Please remember to prepare a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment).

Don't worry about spam. It's okay.
for changelog I thought fragment wasn't needed for new module.  Do i need to do it for proxmox module_utils one?

plugins/modules/proxmox_zone.py
JanaHoch: ahh, I did it in this and few other PRs as well because even for update/create/delete I wanted to verify if it exists or not and I'm calling that get method. if I split it in `zone_info` and `zone` module is it ok if i duplicate those get methods? or let me see if i can create an object of `zone_info` in `zone` and try to access those get methods like  that.

plugins/modules/proxmox_zone.py
JanaHoch: yeah I'm bit wonky on grammar and punctuation. let me fix those.

plugins/modules/proxmox_zone.py
JanaHoch: ok let me check this and update.

plugins/module_utils/proxmox.py
JanaHoch: In README python 3.7 is mentioned so I thought it should be ok. If not let me know i can update but python2 deprecated when the only python I knew was a snake... :)

IamLunchbox: Ah i see, very nice. 

So we could use type hints, as well now, @Thulium-Drake ?

IamLunchbox:
> for changelog I thought fragment wasn't needed for new module. Do i need to do it for proxmox module_utils one?

You are right, the new module should be picked up automatically. Regarding the update of module utils: I, personally, think, that a fragment for these new functions is not needed. They are focused on sdn usage and won't impact other modules. @Thulium-Drake , what do you think?

plugins/modules/proxmox_zone.py
IamLunchbox: Or you could create the information retrieval function in `module_utils`.

I checked the usage of some other examples in there and there are some, which only have one call through their associated module, so two usages in your case should be fine. I don't know if cross-calling functions from two modules is allowed or not. But I suppose, that `module_utils` is exactly the place for these kind of use cases-

plugins/modules/proxmox_zone.py
```diff
@@ -0,0 +1,536 @@
...
+            "reversedns": self.params.get("reversedns"),
+            "rt-import": self.params.get("rt_import"),
+            "tag": self.params.get("tag"),
+            "vlan-protocol": self.params.get("vlan_protocol"),
+            "vrf-vxlan": self.params.get("vrf_vxlan"),
+            "vxlan-port": self.params.get("vxlan_port"),
+        }
+
+        if zone_params['lock-token'] is None and state is not None:
+            zone_params['lock-token'] = self.get_global_sdn_lock()
```
IamLunchbox: This would lock the sdn(s?), even if nothing needs to be changed, right? 

Wouldn't it be better to only acquire a lock after checking that a change is required? I could imagine that large deployments might run into a problem, when two simultaneous tasks run and one crashes due to an existing lock - even though neither task actually needed a lock.

plugins/modules/proxmox_zone.py
```diff
@@ -0,0 +1,536 @@
...
+
+    def get_zones(self, type=None):
+        try:
+            return self.proxmox_api.cluster().sdn().zones().get(type=type)
+        except Exception as e:
+            self.module.fail_json(
+                msg=f'Failed to retrieve zone information from cluster: {e}'
+            )
+
+    def zone_present(self, force, **kwargs):
```
IamLunchbox: This function is not idempotent, in case `force` is true and the type is the same. 

But this issue might be resolved by my next suggestion down below.

plugins/modules/proxmox_zone.py
```diff
@@ -0,0 +1,536 @@
...
+                self.module.exit_json(
+                    changed=True, zone=zone, msg=f'Created new Zone - {zone}'
+                )
+            except Exception as e:
+                self.rollback_sdn_changes_and_release_lock(lock)
+                self.module.fail_json(
+                    msg=f'Failed to create zone {zone}'
+                )
+
+    def zone_update(self, **kwargs):
```
IamLunchbox: `zone_update` and `zone_present` depend strongly onto each other and have some duplicated code.

I suggest only using `state: present|absent` and a separate `update: true` option. Then you can combine the logic of these two functions with each other. And, if i understood your code correctly, using `update: true` is the same as using `force` as an option. So you could possibly reduce complexity and still use the same amount of options. :)

plugins/module_utils/proxmox.py
```diff
@@ -245,3 +245,67 @@ def get_storage_content(self, node, storage, content=None, vmid=None):
...
+            )
+
+    def apply_sdn_changes_and_release_lock(self, lock, release_lock=True):
+        """Apply all SDN changes done under a lock token.
+
+        @param lock: Global SDN lock token
+        @param release_lock: if True release lock after successfully applying changes
```
IamLunchbox: If you indicate parameter types I would suggest using colons, as in the [sphinx style](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html#docstring-sections).

This notation is used by other functions in this file as well. Additionally, at least my code interpreter did not resolve this javadoc parameter style and just printed these type hints as text.

plugins/modules/proxmox_zone.py
```diff
@@ -0,0 +1,536 @@
...
+    )
+
+
+class ProxmoxZoneAnsible(ProxmoxAnsible):
+    def __init__(self, module):
+        super(ProxmoxZoneAnsible, self).__init__(module)
+        self.params = module.params
+
+    def validate_params(self):
+        type = self.params.get('type')
```
IamLunchbox: You redefine the python builtin function `type` . This can lead to unexpected behavior and could give you some headache debugging. For example some downstream code could call `type` to find out the type of a variable and crashes, since a string is not a function to call.

plugins/modules/proxmox_zone.py
```diff
@@ -0,0 +1,536 @@
...
+            try:
+                self.proxmox_api.cluster().sdn().zones().post(**kwargs)
+                self.apply_sdn_changes_and_release_lock(lock)
+                self.module.exit_json(
+                    changed=True, zone=zone, msg=f'Created new Zone - {zone}'
+                )
+            except Exception as e:
+                self.rollback_sdn_changes_and_release_lock(lock)
+                self.module.fail_json(
+                    msg=f'Failed to create zone {zone}'
```
IamLunchbox: I would pass {e} as well here, then the user knows what went wrong.

plugins/modules/proxmox_zone.py
```diff
@@ -0,0 +1,536 @@
...
+
+    def zone_absent(self, zone_name, lock):
+        available_zones = [x.get('zone') for x in self.get_zones()]
+        params = {'lock-token': lock}
+
+        try:
+            if zone_name not in available_zones:
+                self.release_lock(lock)
+                self.module.exit_json(
+                    changed=False, zone=zone_name, msg=f"zone {zone_name} already doesn't exist."
```
IamLunchbox: Feel free to ignore: 
Usually `absent` is the keyword for something being not there, I would use that term here instead of `doesn't exist`.

plugins/module_utils/proxmox.py
JanaHoch: yeah not sure why pycharm generated that doc string params with `@`instead of `:` let me fix those

plugins/modules/proxmox_zone.py
JanaHoch: Ok thanks, I'll split this and other modules and add the get methods in module_utils.

plugins/modules/proxmox_zone.py
JanaHoch: Yeah I was being lazy let me fix that.

plugins/modules/proxmox_zone.py
JanaHoch: I was just checking state and getting lock based on that but sure let me make it so lock is acquired just before post/put calls.

plugins/modules/proxmox_zone.py
JanaHoch: yeah I think I forgot here let me add.

plugins/modules/proxmox_zone.py
JanaHoch: @IamLunchbox / @Thulium-Drake , Do you think it'll be better to create another class in `module_utils` another base class for all SDN utilities. I can add get methods from all other PRs there and all these locking functions. so that It'll be bit more organized? if this sounds stupid let me know I can always keep it in original base class. And of course this new class's parent will be `ProxmoxAnsible`

IamLunchbox: Will several modules, who all deal with sdns/networks use this class or only zone? I did not yet check your other PRs, if they need sdn lock.

JanaHoch: this PR should be ready now. I've created new SDN base class. and fixed all the other things you mentioned and marked them resolved.  except for few which need confirmation. if there is still any change needed let me know.

plugins/module_utils/proxmox.py
Thulium-Drake: Ehm, good question, @felixfontein what are the guidelines in `community.general` for this?

plugins/modules/proxmox_zone.py
Thulium-Drake: Yes, we're still working on it (progress is a bit slow unfortunately), but my plan is that after we get the server up, we'll start working on a molecule setup that will validate some basic modules first. 

And then expand it where we can :-)

Thulium-Drake:
> > for changelog I thought fragment wasn't needed for new module. Do i need to do it for proxmox module_utils one?
> 
> You are right, the new module should be picked up automatically. Regarding the update of module utils: I, personally, think, that a fragment for these new functions is not needed. They are focused on sdn usage and won't impact other modules. @Thulium-Drake , what do you think?

Correct, new modules don't require changelogs as the changelog tool will pick them up itself

Thulium-Drake:
Thanks @JanaHoch and @IamLunchbox for your work! I'm sorry I can't help you more with the intricate details of the python programming. I'm more of a Ansible poweruser and not so much a Python programmer :sweat_smile:, but I'm happy to help where I can!

With regards to the module, the tests pass and you both seem happy with the results ;-)

IamLunchbox:
Hm, I wouldve liked to review the fixed result - I did not know if the changes are all implemented. But what done is done i guess :|

Thulium-Drake:
Well, it's not released yet, I could still revert it if you want. Please let me know how you would like to proceed :-)

plugins/module_utils/proxmox.py
felixfontein: Whether Python 2 compatibility is still needed depends on a) the supported ansible-core versions of the collection, and b) on the collection itself as well (there might be further restrictions on target version numbers).

For community.general, we usually support all ansible-core versions that are not EOL and haven't been EOL for more than a few weeks at the time of a new major release. That means that community.general 11.0.0 (released in spring) still supports ansible-core 2.16, which still supports Python 2.7 on the target. But the next major release, community.general 12.0.0 (should be released in November), will only support ansible-core >= 2.17, and thus will be Python 3.7+.

(I'm not saying this policy is great, but at least it is predictable and doesn't let the CI matrix explode... like community.crypto 2.x.y, which supports everything from Ansible 2.9 up to ansible-core 2.19, or community.internal_test_tools and community.library_inventory_filtering_v1, which even cover Ansible 2.9 up to ansible-core devel :) )

community.proxmox 1.0.0 said it only supports ansible-core 2.17+ from the beginning, so it was always Python 3.7+.

felixfontein:
> > You are right, the new module should be picked up automatically. Regarding the update of module utils: I, personally, think, that a fragment for these new functions is not needed. They are focused on sdn usage and won't impact other modules. @Thulium-Drake , what do you think?
> 
> Correct, new modules don't require changelogs as the changelog tool will pick them up itself

In case anyone wonders on how that works: it looks at the top-level `version_added`. As long as that's there and has the value of the next minor release, the module will be picked up. I just took a quick look, this PR doesn't have `version_added` entries, so the mechanism won't work. So something for a follow-up :)

IamLunchbox:
Thank you for the heads up felix :)

I finished reviewing and created #189 to fix the missing release tag.

JanaHoch:
Thank you all

PR #179, opened 2025-09-04
Add source option for proxmox_tasks_info

fredericgiquel:
##### SUMMARY

Add option to be able to consider all tasks

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

`proxmox_tasks_info`

##### ADDITIONAL INFORMATION

Currently, it's only possible to list archived/finished tasks. As we can see here (https://pve.proxmox.com/pve-docs/api-viewer/index.html#/nodes/{node}/tasks), it's the default behavior of API call when no `source` option is specified.
This PR add the possibility to specify other source (and therefore to list active tasks). The default remains to only consider archived tasks in order to avoid introducing a breaking change.


Thulium-Drake:
Thanks for this work!

Issue #30, opened 2022-05-13
Proxmox SDN support


damo2929:
### Summary

Hi All 
would it be possible to add support for proxmox's SDN implementation also that it would be possible to add and manage networks etc from ansible.

the SDN is described here 
https://pve.proxmox.com/pve-docs/api-viewer/index.html#/cluster/sdn

### Issue Type

Feature Idea

### Component Name

proxmox

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

btravouillon:
@Thulium-Drake I'm not sure #173 completes this feature request. The change helps to get VNET info but don't cover the "add and manage networks" part of this request.

JanaHoch:
Hi @btravouillon ,  for some reason only #173 got tagged to this. but I had raised other PRs that'll create zone, vnet, subnet and all those things. let me know if you still think something is missing.

PR #181, opened 2025-09-06
new_module proxmox_subnet

JanaHoch:
##### SUMMARY

Added new module `proxmox_subnet` to create/update/delete subnets in proxmox SDN

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

proxmox_subnet

##### ADDITIONAL INFORMATION

This is part of #30 
Note: If you are trying to test this. it needs sdn locking functions which I've added in https://github.com/ansible-collections/community.proxmox/pull/176

### DIFF
--- a/plugins/modules/proxmox_subnet.py
+++ b/plugins/modules/proxmox_subnet.py
@@ -24,8 +24,7 @@ attributes:
 options:
   state:
     description:
-      - Desired state of the network configuration.
-      - Choices include present (create), absent (delete), or update (modify).
+      - Desired state of the subnet configuration.
     type: str
     choices: ['present', 'absent']
     default: present


JanaHoch:
Hi @IamLunchbox  can you Please also check this. @Thulium-Drake  this also needs that function from #183

JanaHoch:
Just for reference, when I was thinking through for update method I created this flow chart with the help of GPT to better visualize.  This might not be exactly same as the final code (I created this before writing it)but still it should give general idea of what I was trying.
mermaid-diagram-2x

Thulium-Drake:
@JanaHoch I have updated this one as well with #183 in it! :-)

IamLunchbox:
The flow chart is kind of daunting :laughing: 

But I'll check this PR and #182 out in the immediate future, to not keep the next release waiting! :grin:

JanaHoch:
> The flow chart is kind of daunting 😆
> 
> But I'll check this PR and #182 out in the immediate future, to not keep the next release waiting! 😁

Yeah flow chart looks bit complicated but basically I'm doing checks in 2 stages 
stage 1 - check all other parameters for subnet except dhcp_range. this will tell usif parameters needs to be updated or if this is a complete new subnet

stage 2 - chek dhcp_range. this will tell us if user has supplied any new range which are not already present and if any of them partially overlap with existing one

to decide what to do with dhcp range we have another parameter `dhcp_range_update_mode` 

If there is partial overlap and mode is overwrite we overwrite  if it is append we error out

If no overlap and mode is append/overwrite we append/overwrite dhcp and update other parameters as  needed

If user didn't give any range  and if mode is append we just ignore existing ranges.  or if mode is overwrite we delete with `delete=dhcp_range`. default is always append so it won't accidentally delete.  
let  me know if you think instead of deleting in this it would be just better to warn user or fail and make sure they have to explicitly send `delete=dhcp_range`

plugins/modules/proxmox_subnet.py
```diff
@@ -0,0 +1,433 @@
...
+attributes:
+  check_mode:
+    support: none
+  diff_mode:
+    support: none
+options:
+  state:
+    description:
+      - Desired state of the network configuration.
+      - Choices include present (create), absent (delete), or update (modify).
```
IamLunchbox: Arent these choices self explanatory?

plugins/modules/proxmox_subnet.py
```diff
@@ -0,0 +1,433 @@
...
+      - If set to append and new dhcp_range passed it will just append to existing ranges.
+      - And If no dhcp_range passed and there are existing ranges it will just ignore existing ranges and only update other params if needed
+      - If set to overwrite and new dhcp_range passed it will overwrite existing ranges.
+      - If no dhcp_range passed and there are existing ranges it will delete all dhcp_ranges
+    type: str
+    default: append
+    choices: ['append', 'overwrite']
+  subnet:
+    description:
+      - subnet CIDR.
```
IamLunchbox:
```suggestion
      - Subnet CIDR.
```

plugins/modules/proxmox_subnet.py
```diff
@@ -0,0 +1,433 @@
...
+    type: str
+    required: true
+  vnet:
+    description:
+      - The virtual network to which the subnet belongs.
+    type: str
+    required: true
+  zone:
+    description:
+      - Vnet Zone
```
IamLunchbox:
```suggestion
      - Vnet Zone.
```

plugins/modules/proxmox_subnet.py
```diff
@@ -0,0 +1,433 @@
...
+    vnet: test
+    subnet: 10.10.2.0/24
+    zone: ans1
+    state: present
+    dhcp_range:
+      - start: 10.10.2.5
+        end: 10.10.2.50
+      - start: 10.10.2.100
+        end: 10.10.2.150
+    snat: True
```
IamLunchbox:
```suggestion
    validate_certs: false
    vnet: test
    subnet: 10.10.2.0/24
    zone: ans1
    state: present
    dhcp_range:
      - start: 10.10.2.5
        end: 10.10.2.50
      - start: 10.10.2.100
        end: 10.10.2.150
    snat: true
```

plugins/modules/proxmox_subnet.py
```diff
@@ -0,0 +1,433 @@
...
+
+- name: Delete a subnet
+  community.proxmox.proxmox_subnet:
+    api_user: "{{ pc.proxmox.api_user }}"
+    api_token_id: "{{ pc.proxmox.api_token_id }}"
+    api_token_secret: "{{ vault.proxmox.api_token_secret }}"
+    api_host: "{{ pc.proxmox.api_host }}"
+    validate_certs: no
```
IamLunchbox:
```suggestion
    validate_certs: false
```

IamLunchbox (commented):
LGTM

Thulium-Drake:
@IamLunchbox Thanks for the review!

@JanaHoch Thanks for the work you've done! <3 I will be abroad for a few days, please ping me when you're ready for merging.

When this one is merged, I'd like to do a release as we have soooo much goodies to release right now :partying_face: I hope to find some time to do this on sunday

JanaHoch:
Hi @Thulium-Drake , This is good to go.
Thank you @IamLunchbox , @Thulium-Drake  for all your help. looking forward to release.....

PR #191, opened 2025-09-20
proxmox_vm_info: Do not throw exception when iterating through machines and optional api results are missing

koalajoe23:
##### SUMMARY
Do not access result vars that are documented as optional in Proxmox API Viewer via `var[key]` in proxmox_vm_info as it will raise exceptions during VM enumeration.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
proxmox_vm_info

##### ADDITIONAL INFORMATION
I was using proxmox_vm_info module to dynamically query infos about VMs. Everything worked pretty well while writing this playbook and running manually.

```
    - name: "Check for template present"
      block:
        - name: Check if Template VM is present
          community.proxmox.proxmox_vm_info:
            node: "{{ proxmox_node }}"
            name: "{{ template_vm_name }}"
          register: template_check
          failed_when: template_check.proxmox_vms | length == 0
```

Now I put this ansible playbook into a CI job and it regularly falls apart:

```
TASK [Check if Template VM is present] *****************************************
FAILED - RETRYING: [localhost]: Check if Template VM is present (2 retries left).
FAILED - RETRYING: [localhost]: Check if Template VM is present (1 retries left).
[ERROR]: Task failed: Module failed: Failed to retrieve QEMU VMs information: 'name'
Origin: /workspace/Lehre/hacklab-ansible/CreateVMs.yml:34:11
32     - name: "Check for template present"
33       block:
34         - name: Check if Template VM is present
             ^ column 11
fatal: [localhost]: FAILED! => {"attempts": 2, "changed": false, "failed_when_result": "Error while evaluating conditional: object of type 'dict' has no attribute 'proxmox_vms'", "msg": "Failed to retrieve QEMU VMs information: 'name'"}
```

Root cause is that the current implementation of VM enumeration in proxmox_vm_info expects a VM name to be returned from https://pve.proxmox.com/pve-docs/api-viewer/index.html#/cluster/resources though the name is marked as optional return value, as most VM details.

https://github.com/ansible-collections/community.proxmox/blob/80dc876d8a5242121ef7495b1d1f1fbbb4019d8a/plugins/modules/proxmox_vm_info.py#L177-L182

Because of this, my proxmox_vm_info action fails every time there is any VM on the cluster without a name. I presume this happens for freshly created VMs still in initialization phase.

When testing for optional args `name`, `vmid` and `node` defensively this error does not occur anymore.


Thulium-Drake:
Thanks for this! :rocket:

PR #195, opened 2025-10-05
proxmox: Add delete parameter to delete settings

JanaHoch:
##### SUMMARY

Add delete parameter to delete settings

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

proxmox

##### ADDITIONAL INFORMATION

Currently we can't delete any settings for a vm
For example: when creating LXC I did it with 2 netif 1 which has internet access and another for internal use.
Afterwards I was trying to delete net0 I tried by just passing netif=net1 but that doesn't delete network instead there will be 2 networks with same config.
To fix this we need to send delete=net0 command


Thulium-Drake:
LGTM! Thanks for this! :-)

PR #204, opened 2025-10-23
proxmox_firewall : ipsets

### LINKED ISSUE #198: Add Ipset support to firewall module
##### SUMMARY
Thank you for the new firewall module. I'm implementing it in a role and it seems to work very well for adding aliases, security groups and rules. One feature missing seems to be the ability to create and manage [ip sets](https://pve.proxmox.com/wiki/Firewall#pve_firewall_ip_sets).

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
proxmox_firewall

##### ADDITIONAL INFORMATION
Ip sets allow grouping of hosts and networks making managing firewall rules easier and less verbose.

I imagine the api could look along the lines of the below:

```yml
- name: Create FW ip sets
  community.proxmox.proxmox_firewall:
    api_user: "{{ pc.proxmox.api_user }}"
    api_token_id: "{{ pc.proxmox.api_token_id }}"
    api_token_secret: "{{ vault.proxmox.api_token_secret }}"
    api_host: "{{ pc.proxmox.api_host }}"
    validate_certs: false
    state: present
    ip_sets:
      - name: hypervisors
        comment: PVE hosts
        cidrs: 
          - cidr: 192.168.0.2
            no_match: false
            comment: pve-01
          - cidr: dc/pve2 # alias
            no_match: false
            comment: pve-02
```

JanaHoch:
##### SUMMARY

Add ipset support for firewall

This fixes #198 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

proxmox_firewall
proxmox_firewall_info

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/proxmox_firewall_info.py
+++ b/plugins/modules/proxmox_firewall_info.py
@@ -87,7 +87,7 @@ groups:
 ip_sets:
     description:
       - List of IP Sets.
-      - These are only supported on the O(cluster) level, other inputs are ignored.
+      - These are only supported on the O(level) = cluster, other inputs are ignored.
     returned: on success
     type: list
     elements: dict


JanaHoch:
Hi @Thulium-Drake — I’ll add unit tests today or tomorrow and I’m working on one more PR for firewall config options. Do you want to do the initial 1.4.0 release before these changes, or include them in the following release? I’ll add changelog fragments if we release 1.4.0 now; if we delay and bundle everything into the new module, we can skip those fragments.

plugins/module_utils/proxmox_sdn.py
```diff
@@ -152,7 +152,8 @@ def get_ip_sets(self):
         """
         ip_sets = self.proxmox_api.cluster().firewall().ipset().get()
         for ip_set in ip_sets:
-            cidrs = self.proxmox_api.cluster().firewall().ipset(ip_set['name']).get()
+            ip_set_obj = getattr(self.proxmox_api.cluster().firewall().ipset(), ip_set['name'])
+            cidrs = ip_set_obj.get()
```
JanaHoch: This was done for unit test. without this magic mock was failing.

Thulium-Drake:
I didn't have time in the last few days to make the release. I can wait on this one as the firewall module has not (yet) been released. So we release it in a more complete fashion ;-)

Please ping me when you think it's ready!

JanaHoch:
> I didn't have time in the last few days to make the release. I can wait on this one as the firewall module has not (yet) been released. So we release it in a more complete fashion ;-)
> 
> Please ping me when you think it's ready!

This is ready and @Freddy500 tested it. 

but for the other PR for firewall config options I'm facing one challenge. currently proxmox api doesn't send back values for parameters which were not modified after creation. [forum post](https://forum.proxmox.com/threads/proxmox-http-api-firewall-doesnt-return-any-optional-values.88636/) do you think we should only display the values we are getting from API or should we keep a map for all the default proxmox values and when anyone runs `info` module we will check if API is returning those if not then just return the default values.

Thulium-Drake:
Oof, that's unfortunate, but from what I gathered from the (age of the) forum post, they don't seem to be planning to address this in the API.

How difficult would it be to make it work with these steps:

- Retrieve default settings from docs endpoint (that should be present on each node as well)
- Retrieve the changed settings and layer them over the defaults
- Print that to the user

The main reasoning behind this would be, that if defaults change, the module will not produce false information.

That said, I _think_ that could be considered an enhancement to the module, as the API does not provide this functionality either. And I don't think it's worth staying the release of the module for now.

If you agree, I think it's good to create a new issue and work on it Soon(tm)

JanaHoch:
> Oof, that's unfortunate, but from what I gathered from the (age of the) forum post, they don't seem to be planning to address this in the API.
> 
> How difficult would it be to make it work with these steps:
> 
>     * Retrieve default settings from docs endpoint (that should be present on each node as well)
> 
>     * Retrieve the changed settings and layer them over the defaults
> 
>     * Print that to the user
> 
> 
> The main reasoning behind this would be, that if defaults change, the module will not produce false information.
> 
> That said, I _think_ that could be considered an enhancement to the module, as the API does not provide this functionality either. And I don't think it's worth staying the release of the module for now.
> 
> If you agree, I think it's good to create a new issue and work on it Soon(tm)

Yes I agree. I'll slowly start the work to check how to extract this from doc API and all that. But it will definitely take more time. so let's release that with later version. for now let's just focus on this ipset. let me know if you see any issues with implementation/Documentation for this.  Thanks :)

JanaHoch:
even in the docs they have not given default value for all parameters [doc](https://pve.proxmox.com/pve-docs/chapter-pve-firewall.html#pve_firewall_cluster_wide_setup)  :(

I'm not sure if this is intended or another bug. But there is no default value for `enable` not in the doc and not even in their git repo for [apidata](https://git.proxmox.com/?p=pve-docs.git;a=blob;f=api-viewer/apidata.js#l5933) same for [policy_in](https://git.proxmox.com/?p=pve-docs.git;a=blob;f=api-viewer/apidata.js#l5976), [policy_forward](https://git.proxmox.com/?p=pve-docs.git;a=blob;f=api-viewer/apidata.js#l5967), [policy_out](https://git.proxmox.com/?p=pve-docs.git;a=blob;f=api-viewer/apidata.js#l5986). And these can have real impact.....

JanaHoch:
HI, Just to keep this focused on IPset and to get a wider audience for firewall config issue. I've created  separate issue for it. we can track it there.

plugins/modules/proxmox_firewall.py
```diff
@@ -483,13 +596,105 @@ def run(self):
...
+                    existing_ip_set_cidrs = [x['cidrs'] for x in existing_ip_sets if x['name'] == ip_set_name][0]
+                    cidrs_to_create, cidrs_to_update = compare_list_of_dicts(
+                        existing_list=existing_ip_set_cidrs,
+                        new_list=ip_set['cidrs'],
+                        uid='cidr',
+                        params_to_ignore=['digest'],
+                    )
+
+                if cidrs_to_update and not update:
+                    self.module.fail_json(f'Ip set {ip_set_name} needs to be updated but update is false.')
```
Thulium-Drake:
```suggestion
                    self.module.fail_json(f'IP set {ip_set_name} needs to be updated but update is false.')
```

plugins/modules/proxmox_firewall_info.py
```diff
@@ -84,6 +84,55 @@
     sample:
       [ "test" ]
 
+ip_sets:
+    description:
+      - List of IP Sets
```
Thulium-Drake:
```suggestion
      - List of IP Sets.
```

plugins/modules/proxmox_firewall_info.py
```diff
@@ -84,6 +84,55 @@
...
+      - This will always be given at cluster level regardless of the level passed.
```
Thulium-Drake:
```suggestion
      - These are only supported on the O(cluster) level, other inputs are ignored.
```

plugins/modules/proxmox_firewall_info.py
```diff
@@ -84,6 +84,55 @@
...
+      - because only at cluster level we can have IP Sets.
```
Thulium-Drake:
```suggestion
```

Thulium-Drake:
Some minor text changes, LGTM otherwise :-)

PR #218, opened 2025-11-10
Add timeout as parameter to promoxer obejct creation

### LINKED ISSUE #160: Proxmox API timeout on task verification while rollback
##### SUMMARY

When doing community.general.rollback on a slow PVE node, I get the following error : 
`Unable to retrieve API task ID from node [XXX]: HTTPSConnectionPool(host='XXX.fr', port=8006): Read timed out. (read timeout=5)`

This is despite my rollback having a timeout of 120s, it still timeouts at 5s.

The issue is in proxmox_snap.py : 
```
    def snapshot_rollback(self, vm, vmid, timeout, snapname):
        if self.module.check_mode:
            return True

        taskid = self.snapshot(vm, vmid)(snapname).post("rollback")
        while timeout >= 0:
            if self.api_task_ok(vm['node'], taskid):
                return True
            if timeout == 0:
                self.module.fail_json(msg='Reached timeout while waiting for rolling back VM snapshot. Last line in task before timeout: %s' %
                                      self.proxmox_api.nodes(vm['node']).tasks(taskid).log.get()[:1])

            time.sleep(1)
            timeout -= 1
        return False
```
if fails on `self.api_task_ok(vm['node'], taskid)`, because reading the task status timeout after 5 seconds (default proxmoxer timeout). And my server is relatively slow, so it takes (sometimes) more than 5 seconds to answer the API.

Unfortunately, proxmoxer timeout is defined when calling ProxmoxAPI(), and you cannot override it later. So to solves this, we can either : 
1. Introduces a new optional parameter `api_timeout` which is added to ProxmoxAPI() in `ProxmoxAnsible::_connect`, which control the API timeout, with a warning that an high value for this parameter could result in very slow operations
2. Same as 1., but instead of using a new parameter, just use the existing `timeout` parameter
3. Change the logic in self.api_task_ok, and probably other functions, to handle the `timeout` parameter by catching the ReadTimeoutException, and re-trying the request until either it's OK or timeout is reached.

Imo, option 3 will be error-prone, since we need to update basically all function using the API (and who can timeout). I prefer option 1 over option 2, because in option 2 we are changing the functionality of an existing parameter, so it might break existing code (I don't have any example but i prefer to be cautious).

I'm happy to provide a PR if option 1 seems good for you.

qwertiko:
##### SUMMARY
The timeout parameter has been added to the creation of the proxmoxer object, because the timeout parameter via the ansible module (default: 30) has no effect on some functions. 

Fixes #160

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
modules_utils/proxmox.py 

##### ADDITIONAL INFORMATION

If the Proxmox object is created as usual, the default timeout is 5 seconds. However, if there are a relatively large number of objects in the storage, the Proxmox API may take longer than 5 seconds to respond to the content_check function request (module/proxmox.py).

Result: container creation fails.


Thulium-Drake:
Thanks for this! Can you please add a changelog fragment? :-)

qwertiko:
Change log fragment added. 

I added also an if-clause and a default value because I noticed that modules such as proxmox_firewall do not have a timeout flag at all. If these are executed without a var check, the code crashes because a variable is to be used that does not exist.

plugins/module_utils/proxmox.py
```diff
@@ -148,6 +149,11 @@ def _connect(self):
         api_token_secret = self.module.params['api_token_secret']
         validate_certs = self.module.params['validate_certs']
 
+        if 'timeout' in self.module.params:
+            timeout = self.module.params['timeout']
+        else:
```
Thulium-Drake:
```suggestion
        else:
```

plugins/module_utils/proxmox.py
```diff
@@ -148,6 +149,11 @@ def _connect(self):
         api_token_secret = self.module.params['api_token_secret']
         validate_certs = self.module.params['validate_certs']
 
+        if 'timeout' in self.module.params:
+            timeout = self.module.params['timeout']
+        else: 
+            timeout = PROXMOXER_DEFAULT_TIMEOUT
+
```
Thulium-Drake:
```suggestion
```

Thulium-Drake (commented):
Fixed some whitespace issues

PR #219, opened 2025-11-12
Add ignore.txt entries for new validate-modules check

felixfontein:
##### SUMMARY
CI currently fails: https://github.com/ansible-collections/community.proxmox/actions/runs/19289866683/job/55157926361

Ref: https://forum.ansible.com/t/44833

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
CI


felixfontein:
BTW, I'm still receiving mails if the nightly CI breaks for some reason. The problem is that GitHub always sends them to the last person who activated these tests (which apparently was me). You can change this by going to https://github.com/ansible-collections/community.proxmox/actions/workflows/nox.yml, click on the three-dots menu in the upper right corner, select "Disable workflow", and then re-enable it.

Thulium-Drake:
Thanks for the fix :-) I've kicked the workflow (so they should come back to me now)

felixfontein:
Thanks for reviewing and merging! You're also the better address of sending the mails to, if it goes through me every time it can be quite a waste of time ;)

PR #226, opened 2025-11-22
new_modules_ceph_mon mgr and mds

teslamania31:
##### SUMMARY
Modules to add or delete ceph manager, monitors and metadata service

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
- proxmox_ceph_mds
- proxmox_ceph_mgr
- proxmox_ceph_mon

PR #227, opened 2025-11-23
new module proxmox_sendkey

### LINKED ISSUE #47: Add support proxmox sendkeys function
### Summary

Proxmox api has `sendkey` function.

https://pve.proxmox.com/pve-docs/api-viewer/index.html#/nodes/{node}/qemu/{vmid}/sendkey

This function is useful for VM initial installation by Ansible.

So, it would be better to create that.

we can used as reference `vmware_guest_sendkey`.

https://docs.ansible.com/ansible/latest/collections/community/vmware/vmware_guest_sendkey_module.html

### Issue Type

Feature Idea

### Component Name

proxmox_sendkey

### Additional Information

```yaml (paste below)

```

### Code of Conduct

miyuk:
##### SUMMARY

New modules for sendkey to VM

Fixes #47 

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

proxmox_sendkey

##### ADDITIONAL INFORMATION

Example
```yaml
- name: Send Ctrl+Alt+Delete to a Windows VM
  proxmox_sendkey:
    api_host: "{{ proxmox_host }}"
    api_user: "{{ proxmox_user }}"
    api_password: "{{ proxmox_password }}"
    name: win-test
    keys_send:
      - ctrl-alt-delete
```

### DIFF
--- a/plugins/modules/proxmox_sendkey.py
+++ b/plugins/modules/proxmox_sendkey.py
@@ -42,14 +42,15 @@ options:
       - List of keys or key sequence to send in order.
       - Each item must follow the qemu key naming format such as
         C(ctrl-alt-delete) or C(ret).
-      - You can specify either O(keys_send) or O(string_send) of them.
+      - You can specify either O(keys_send) or O(string_send).
     type: list
     elements: str
   string_send:
     description:
       - Raw string that will be transformed to the corresponding key presses
+      - Only ASCII-characters are supported.
         before sending.
-      - You can specify either O(keys_send) or O(string_send) of them.
+      - You can specify either O(keys_send) or O(string_send).
     type: str
   delay:
     description:
@@ -425,7 +426,7 @@ class ProxmoxSendkeyAnsible(ProxmoxAnsible):
         for ch in str(text):
             ch_keys = self.CHAR_MAP.get(ch, None)
             if ch_keys is None:
-                self.module.fail_json(f"string char is invalid: {ch}")
+                self.module.fail_json(f"The character {ch} passed to string_send is not allowed")
             key = "-".join(ch_keys)
             keys.append(key)
         return keys


plugins/modules/proxmox_sendkey.py
```diff
@@ -0,0 +1,481 @@
...
+attributes:
+  check_mode:
+    support: none
+  diff_mode:
+    support: none
+notes:
+  - Keys must match the qemu key names listed in
+    L(QEMU ui.QKeyCode,https://www.qemu.org/docs/master/interop/qemu-qmp-ref.html#enum-QMP-ui.QKeyCode).
+options:
+  name:
```
Thulium-Drake, line 31: This seems wrong, is this also the `name` of the VM? Or the name of the keystroke sequence?

plugins/modules/proxmox_sendkey.py
```diff
@@ -0,0 +1,481 @@
...
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+DOCUMENTATION = r"""
+---
+module: proxmox_sendkey
+short_description: Send key presses to a Proxmox VM console
+version_added: "1.4.0" # TODO
```
Thulium-Drake:
```suggestion
version_added: 1.5.0
```

Thulium-Drake (changes requested):
LGTM, some nitpicking on the documentation in the module

Thulium-Drake:
@IamLunchbox / @JanaHoch Can one of you please take a look at this? If you're happy, I'm happy to merge it! :-)

IamLunchbox:
I will check it out :)

plugins/modules/proxmox_sendkey.py
miyuk: Sorry, I don't understand what you're pointing out.

This means **name of the VM**.
I added the option because other module contains this options.
ex.) https://github.com/ansible-collections/community.proxmox/blob/fee999c640d23077bcb9774c1c3bce4806ee86a2/plugins/modules/proxmox_vm_info.py#L37

Thulium-Drake: Oh, right, durr, I wasn't paying attention, you're using this to target the VM :sweat_smile:

IamLunchbox:
Just FYI: I'll have time tomorrow to check it out.

plugins/modules/proxmox_sendkey.py
```diff
@@ -0,0 +1,481 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+#
+# Copyright (c) 2025, (@teslamania) <nicolas.vial@protonmail.com>
```
IamLunchbox: Do you have 2 accounts or did you copy the copyright from #226?

plugins/modules/proxmox_sendkey.py
```diff
@@ -0,0 +1,481 @@
...
+options:
+  name:
+    description:
+      - The unique name of the VM.
+      - You can specify either O(name) or O(vmid) or both of them.
+    type: str
+  vmid:
+    description:
+      - The unique ID of the VM.
+      - You can specify either O(vmid) or O(name) or both of them.
```
IamLunchbox: Either tell users, that one does replace the other or - what i'd prefer - make them mutually exclusive.

```suggestion
  vmid:
    description:
      - The unique ID of the VM.
      - Takes precedence over O(name).
```

plugins/modules/proxmox_sendkey.py
```diff
@@ -0,0 +1,481 @@
...
+  diff_mode:
+    support: none
+notes:
+  - Keys must match the qemu key names listed in
+    L(QEMU ui.QKeyCode,https://www.qemu.org/docs/master/interop/qemu-qmp-ref.html#enum-QMP-ui.QKeyCode).
+options:
+  name:
+    description:
+      - The unique name of the VM.
+      - You can specify either O(name) or O(vmid) or both of them.
```
IamLunchbox:
```suggestion
```

plugins/modules/proxmox_sendkey.py
```diff
@@ -0,0 +1,481 @@
...
+    type: str
+  vmid:
+    description:
+      - The unique ID of the VM.
+      - You can specify either O(vmid) or O(name) or both of them.
+    type: int
+  keys_send:
+    description:
+      - List of keys or key chords to send in order.
```
IamLunchbox:
```suggestion
      - List of keys or key sequence to send in order.
```

plugins/modules/proxmox_sendkey.py
```diff
@@ -0,0 +1,481 @@
...
+      - Each item must follow the qemu key naming format such as
+        C(ctrl-alt-delete) or C(ret).
+      - You can specify either O(keys_send) or O(string_send) or both of them.
+    type: list
+    elements: str
+  string_send:
+    description:
+      - Raw string that will be transformed to the corresponding key presses
+        before sending.
+      - You can specify either O(string_send) or O(keys_send) or both of them.
```
IamLunchbox: But `string_send` takes precedence over `keys_send`, at least in your code. Your wording implies that both will be executed - and neither does that happen nor does the user know, which executes first.

I'd make them mutually exclusive.

plugins/modules/proxmox_sendkey.py
```diff
@@ -0,0 +1,481 @@
...
+        self.total_keys = []
+        self.completed_keys = []
+
+    def string_to_keys(self, text):
+        """Convert text to key list."""
+        keys = []
+        for ch in str(text):
+            key = "-".join(self.CHAR_MAP[ch])
+            keys.append(key)
+        return keys
```
IamLunchbox, line 432: This will crash if a not-expected character, as `ö` or some such, is passed to `send_string`. Since users might enter several unsupported unicode characters, I'd catch these and give an appropriate error.

My expecation is, that this otherwise raises a KeyError and leaves users in the weeds. You have a validation function, but that would only catch errors in malformed keys passed in as `keys_send`.

plugins/modules/proxmox_sendkey.py
```diff
@@ -0,0 +1,481 @@
...
+        if not vmid:
+            vmid = self.get_vmid(name)
+
+        # Convert text to key list
+        if string_send:
+            keys_send = self.string_to_keys(string_send)
+        self.total_keys = keys_send
+        self.validate_keys(self.total_keys)
+
+        self.send_keys(vmid, self.total_keys, dalay)
```
IamLunchbox:
```suggestion
        delay = self.params.get("delay")

        # Get vmid from name
        if not vmid:
            vmid = self.get_vmid(name)

        # Convert text to key list
        if string_send:
            keys_send = self.string_to_keys(string_send)
        self.total_keys = keys_send
        self.validate_keys(self.total_keys)

        self.send_keys(vmid, self.total_keys, delay)
```

IamLunchbox (changes requested):
I found some logical errors, which should be cleared up before merging. In my opinion letting users enter different options simultaneously needs to document for them, which option is heeded when both are given and valid. 

I personally would prefer having mutually exclusive options for `vmid` and `name` as well as `send_key` and `string_send`.

plugins/modules/proxmox_sendkey.py
miyuk: Excuse me, I copied from #226.
I will fix it.

plugins/modules/proxmox_sendkey.py
miyuk: I added charactor check in this function.
How about it?

plugins/modules/proxmox_sendkey.py
miyuk: Add `mutually exclusive` option

plugins/modules/proxmox_sendkey.py
```diff
@@ -0,0 +1,486 @@
...
+    description:
+      - The unique ID of the VM.
+      - Takes precedence over O(name).
+    type: int
+  keys_send:
+    description:
+      - List of keys or key sequence to send in order.
+      - Each item must follow the qemu key naming format such as
+        C(ctrl-alt-delete) or C(ret).
+      - You can specify either O(keys_send) or O(string_send) of them.
```
IamLunchbox:
```suggestion
      - You can specify either O(keys_send) or O(string_send).
```

plugins/modules/proxmox_sendkey.py
```diff
@@ -0,0 +1,486 @@
...
+    type: list
+    elements: str
+  string_send:
+    description:
+      - Raw string that will be transformed to the corresponding key presses
+        before sending.
+      - You can specify either O(keys_send) or O(string_send) of them.
```
IamLunchbox:
```suggestion
      - You can specify either O(keys_send) or O(string_send).
```

plugins/modules/proxmox_sendkey.py
```diff
@@ -0,0 +1,486 @@
...
+        self.total_keys = []
+        self.completed_keys = []
+
+    def string_to_keys(self, text):
+        """Convert text to key list."""
+        keys = []
+        for ch in str(text):
+            ch_keys = self.CHAR_MAP.get(ch, None)
+            if ch_keys is None:
+                self.module.fail_json(f"string char is invalid: {ch}")
```
IamLunchbox:
```suggestion
                self.module.fail_json(f"The character {ch} passed to string_send is not allowed")
```

plugins/modules/proxmox_sendkey.py
```diff
@@ -0,0 +1,486 @@
...
+    description:
+      - List of keys or key sequence to send in order.
+      - Each item must follow the qemu key naming format such as
+        C(ctrl-alt-delete) or C(ret).
+      - You can specify either O(keys_send) or O(string_send) of them.
+    type: list
+    elements: str
+  string_send:
+    description:
+      - Raw string that will be transformed to the corresponding key presses
```
IamLunchbox, line 50:
```suggestion
      - Raw string that will be transformed to the corresponding key presses
      - Only ASCII-characters are supported.
```
Or some such, looked similar to the ascii keyspace to me. **But I did not take a deeper look into it**

Thulium-Drake:
@miyuk Thanks for your work!

@IamLunchbox Thanks for the reviews! 

:rocket:

PR #248, opened 2026-01-01
Fix ipset_bugs in proxmox_firewall

IamLunchbox:
This PR fixes three shortcomings of `proxmox_firewall`:

Without this patch, `vm` level ipsets will be created on cluster level. To fix that, `firewall_obj` is initialized at the respective level and then passed down to the ipset functions, enabling either cluster or vm-level changes.

Proxmox rewrites the entered ipsets to use the correct base ip of the cidr, e.g. `10.0.0.11/24` will be `10.0.0.0/24` for proxmox. Adding ipsets will work once this way, adding an ipset a second time failed in the past and it was not idempotent either. This PR changes the user data to be in the expected format for proxmox.

Additionally: Proxmox changes /32 ipsets to be a single ip without netmask. This gets fixed as well.


Thulium-Drake:
@IamLunchbox Awesome work thanks! It LGTM, do you want @JanaHoch to give it a spin before I merge it?

Issue #71, opened 2024-11-19
proxmox_template upload timeout=5s can't be changed and too short


elelayan:
### Summary

When I upload an iso to my cluster it takes ~20s.

See  is the upload task.

Here is the error I get:

````
msg: 'Uploading template files/isos/myhost.iso failed with error: (''Connection aborted.'', TimeoutError(''The write operation timed out''))'
````

I think the timeout needs to be specified when creating the ProxmoxAPI object in [proxmox.py](https://github.com/ansible-collections/community.general/blob/main/plugins/module_utils/proxmox.py#L109).

At least it works for me when I change to `return ProxmoxAPI(api_host, verify_ssl=validate_certs, timeout=42, **auth_args)`.
And I see the effect when logging the timeout value in `ProxmoxHttpSession` in `proxmoxer/backend/https.py` line 232.

### Issue Type

Bug Report

### Component Name

proxmox_template

### Ansible Version

```console (paste below)
$ ansible --version
ansible [core 2.14.17]
  config file = /home/xxx/.../ansible.cfg
  configured module search path = ['/home/xxx/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/xxx/.tox/py3-ansible/lib/python3.11/site-packages/ansible
  ansible collection location = /home/xxxxx/collections-dependencies:/home/xxxxx/collections
  executable location = /home/xxxx/.tox/py3-ansible/bin/ansible
  python version = 3.11.2 (main, Aug 26 2024, 07:20:54) [GCC 12.2.0] (/home/xxxxx/.tox/py3-ansible/bin/python)
  jinja version = 3.1.4
  libyaml = True
```

### Community.general Version

```console (paste below)
$ ansible-galaxy collection list community.general

# /home/xxxx/collections-dependencies/ansible_collections
Collection        Version
----------------- -------
community.general 8.6.7  
```

### Configuration

```console (paste below)
$ ansible-config dump --only-changed

```

### OS / Environment

Debian 12

### Steps to Reproduce

```yaml (paste below)
    - name: Upload CDROM
      community.general.proxmox_template: "{{ pve_api | combine(task_params) }}"
      delegate_to: localhost
      vars:
        task_params:
          content_type: iso
          src: "files/isos/{{ inventory_hostname }}.iso"
          storage: "{{ pve_iso_storage }}"
          force: true
```

### Expected Results

I expected the upload to succeed.

### Actual Results

```console (paste below)
[...]

Issue #55, opened 2024-02-25
proxmox_kvm: Set timeout on Proxmoxer connections


aph3rson:
### Summary

This is related to [this PVE Forums thread](https://forum.proxmox.com/threads/intermittent-network-loss-from-vm-node.142042/). I'm having issues with a slow-to-respond PVE API, and the current timeout is set to the default of 5 seconds. I'd like to be able to increase that.

### Issue Type

Feature Idea

### Component Name

proxmox_kvm

### Additional Information

Here's an example of what this might look like in a template:

```yaml
      community.general.proxmox_kvm:
        api_host: "{{ hostvars.pve.ansible_host }}"
        api_user: "{{ hostvars.pve.ansible_user }}@pam"
        api_password: "{{ hostvars.pve.ansible_ssh_pass }}"
        api_timeout: 10  # <--- new
        node: pve-1
        vmid: "{{ item.value.vmid }}"
        state: absent
```

### Code of Conduct

- [X] I agree to follow the Ansible Code of Conduct

PR #253, opened 2026-01-12
Add a global api timeout option

### LINKED ISSUE #233: Add timeout param to proxmox_vm_info module.
##### SUMMARY

Add timeout param to proxmox_vm_info module.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

proxmox_vm_info

##### ADDITIONAL INFORMATION

On Proxmox hosts with a significant number of VMs, the /ap2/json/nodes/NODENAME/qemu call the modules makes can take longer than 5 seconds to respond. Proxmoxer's default timeout is 5 seconds, causing these requests to fail.

https://github.com/ansible-collections/community.proxmox/pull/218 added some support for passing a timeout value through to Proxmoxer, but the proxmox_vm_info module doesn't allow the timeout param currently.

```yaml
        - name: "Default or no Timeout - Look for VM Base Template on {{ inventory_hostname }}"
          community.proxmox.proxmox_vm_info:
            api_host: "{{ inventory_hostname }}"
            api_user: "{{ pve_user }}"
            api_token_id: "{{ pve_token_id | default(omit) }}"
            api_token_secret: "{{ pve_token_secret | default(omit) }}"
            name: "{{ default_template_vm_name }}"
          register: vmbase_template_info
          failed_when: not vmbase_template_info.proxmox_vms
        - name: "With Timeout - Look for VM Base Template on {{ inventory_hostname }}"
          community.proxmox.proxmox_vm_info:
            api_host: "{{ inventory_hostname }}"
            api_user: "{{ pve_user }}"
            api_token_id: "{{ pve_token_id | default(omit) }}"
            api_token_secret: "{{ pve_token_secret | default(omit) }}"
            name: "{{ default_template_vm_name }}"
            timeout: 30
          register: vmbase_template_info
          failed_when: not vmbase_template_info.proxmox_vms
```

IamLunchbox:
This PR extends the work from  #218 and introduces the `api_timeout` option for all modules. #218 only selectively enabled modules to use a timeout parameter or setting the option through an environment variable when using api connections.

With this extention of the argument spec all modules using the proxmox api have the ability to request an api timeout to quit early or extend the runtime, if the proxmox api is slow to respond. This change should not change current behaviour, since the default value for proxmoxer, 5 seconds, is kept as a default value.

And other modules periodically checking tasks usually use the timeout differently by querying the api periodically within the timeout limit.

Fixes #233, #71, #55, #45.

Furthermore #218 derived the connect timeout from other modules' `timeout` options, while the authors of these modules might not have anticipated that. For example a user might set a timeout in `proxmox_kvm` of 30 seconds, but would not want to wait 30 seconds for a failed http request. This PR reverts that behavour from #218 by creating a dedicated parameter. 

If users relied on this specific behaviour this could be considered a breaking change. But `timeout` as a module parameter is seldomly present, #218  is rather new and this specific change was never actively documented besides the changelogs. So I suppose it is more a bugfix then a breaking change.

### DIFF
--- a/antsibull-nox.toml
+++ b/antsibull-nox.toml
@@ -13,6 +13,8 @@
 [sessions.docs_check]
 validate_collection_refs="all"
 
+[sessions.license_check]
+
 [sessions.extra_checks]
 run_no_unwanted_files = true
 no_unwanted_files_module_extensions = [".py"]


Thulium-Drake:
I agree with your reasoning, the `timeout` was still rather new and this PR actually solves the issue amongst all modules, which means it's a better fix :-)

So, thank you! I'll try to work through the PRs in the coming week or 2 and then make a new release.

PR #254, opened 2026-01-17
make sdn modules compatible with pve8

### LINKED ISSUE #252: Be able to use /cluster/sdn related modules with pve8
##### SUMMARY
Be able to use `/cluster/sdn` related modules with pve8 by skipping to acquire the global lock for sdn configuration which is not available in pve8

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
proxmox.zone, proxmox_vnet, proxmox_subnet

##### ADDITIONAL INFORMATION
The current modules which are using the `/cluster/sdn` api are using `/cluster/sdn/lock` to acquire the global lock for the sdn configuration. This endpoint is only available in pve9 but not in pve8 and therefore the modules fails when running  aganst the pve8 api.
I would propose to check the current proxmox version and only call the lock api if the version >=9.

florianfischer91:
##### SUMMARY
Don't call `cluster/sdn/{lock,rolback}` api endpoints when proxmox version is < 9, because those endpoints are only available > 9.
Additionally wait for the network load task to complete otherwise multiple requests at the same time will fail in proxmox.

Fixes #252 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
proxmox_sdn, proxmox_zone, proxmox_vnet, proxmox_subnet

### DIFF
--- a/changelogs/fragments/252-make-sdn-modules-compatible-with-pve8.yaml
+++ b/changelogs/fragments/252-make-sdn-modules-compatible-with-pve8.yaml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+    - proxmox_zone, proxmox_vnet, proxmox_subnet - make sdn modules compatible with pve8 (https://github.com/ansible-collections/community.proxmox/pull/254).


microtechno9000:
Keen to see this approved, causing issues with system setup scripts

IamLunchbox:
I don't when the next release is scheduled, you might need to install from git as soon as this PR is merged. 

@Thulium-Drake is able to merge, but I don't think he reviewed the code yet.

Thulium-Drake:
Thanks for this work! :-)

Issue #62, opened 2024-07-14
proxmox*: add parameter for specifying CA cert path


RussellAult:
### Summary

Currently the various `community.general.proxmox*` plugins have two documented options for dealing with TLS certs: using the PVE-generated certs and ignoring that they aren't trusted (the default), or installing trusted certs on the nodes and setting `validate_certs` to `true`. Proxmox itself supports a third option: [importing PVE's root CA certificate into the local trusted certs inventory and using it to verify the PVE-generated certs](https://pve.proxmox.com/wiki/Import_certificate_in_browser). Taking this approach with the `community.general.proxmox*` plugins is made complicated by the fact that the Python `requests` module (which is what `proxmoxer` uses to handle the actual API calls) doesn't rely on the system's `ca-certificates` bundle (i.e. simply importing PVE's root CA cert into the system bundle won't make `requests` trust it).

My current work-around is to set `environment: REQUESTS_CA_BUNDLE: "/path/to/pve-root-ca.pem"` (in addition to `verify_certs: true`) for all `community.general.proxmox*` tasks, but there might be a cleaner way using the `validate_certs` option itself. Currently, the value for this option is simply passed verbatim to `proxmoxer`'s `verify_ssl` parameter, and `proxmoxer` in turn simply passes `verify_ssl` to `requests`'s `verify` parameter, which accepts a path to a CA cert in addition to the usual boolean values. In theory this means it should be possible to provide `community.general.proxmox*`'s `verify_certs` with the path to a CA cert and have `requests` verify any HTTPS connections against the specified cert (and with minimal code changes to boot).

### Issue Type

Feature Idea

### Component Name

proxmox

### Additional Information

```yaml (paste below)
- name: Current way to verify HTTPS connections against PVE's root CA
  community.general.proxmox_kvm:
    api_user: "{{ proxmox_api_user }}"
    api_token_id: "{{ proxmox_api_token_id }}"
    api_token_secret: "{{ proxmox_api_token_secret }}"
    api_host: "{{ proxmox_api_host }}"
    name: "{{ hostvars[inventory_hostname]['inventory_hostname_short'] }}"
    state: started
    validate_certs: true
  delegate_to: localhost
  environment: "{{ omit if proxmox_ca_path is not defined else {'REQUESTS_CA_BUNDLE': proxmox_ca_path} }}"

- name: Proposed way to verify HTTPS connections against PVE's root CA
  community.general.proxmox_kvm:
[...]

PR #256, opened 2026-01-17
Add global option to specify an alternate ca path

IamLunchbox:
This PR adds a global option to specify a ca-certificate used for tls-validation. It uses the `verify` option of [requests](https://docs.python-requests.org/en/latest/user/advanced/#ssl-cert-verification) to specify a certificate, which should be used for tls validation. 

Since proxmoxer passes the `verify_ssl` kwarg down to requests `verify`, passing the certificate down to proxmoxer will in turn lead to successfull connections with requests.

I am not sure if the configuration title of `ca_path` is the best choice. :shrug: 

This change might be breaking though, if we actually really change the default tls validation from `False` to  `True`. I think this would be wise from a security point of view, but it would definetely impact existing users and playbooks, since most will not use valid certificates on their proxmox-apis.

### DIFF
--- a/plugins/doc_fragments/proxmox.py
+++ b/plugins/doc_fragments/proxmox.py
@@ -51,16 +51,15 @@ options:
     type: path
   api_timeout:
     description:
-      - Time limit for requests towards the Proxmox api.
+      - Time limit for requests towards the Proxmox VE API.
     type: int
     default: 5
   validate_certs:
     description:
-      - If V(false), SSL certificates will not be validated.
-      - This should only be used on personally controlled sites using self-signed certificates.
+      - Validate the TLS certificates used for the connection to the Proxmox VE API.  
+      - Currently defaults to V(false) and changes default to V(true) with community.proxmox 2.0.0.
       - Uses the E(PROXMOX_VALIDATE_CERTS) environment variable if not specified.
     type: bool
-    default: false
 requirements: ["proxmoxer >= 2.0", "requests"]
 """
 
--- a/plugins/doc_fragments/proxmox.py
+++ b/plugins/doc_fragments/proxmox.py
@@ -46,16 +46,16 @@ options:
     type: str
   ca_path:
     description:
-      - Path to a local certificate, which will be used to verify tls connections.
+      - Path to a local certificate, which will be used to verify TLS connections.
       - Ignored if O(validate_certs=false).
-    type: str
+    type: path
   validate_certs:
     description:
       - If V(false), SSL certificates will not be validated.
       - This should only be used on personally controlled sites using self-signed certificates.
       - Uses the E(PROXMOX_VALIDATE_CERTS) environment variable if not specified.
     type: bool
-    default: true
+    default: false
 requirements: ["proxmoxer >= 2.0", "requests"]
 """
 
--- a/plugins/doc_fragments/proxmox.py
+++ b/plugins/doc_fragments/proxmox.py
@@ -2,11 +2,10 @@
 # GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
-from __future__ import (absolute_import, division, print_function)
-__metaclass__ = type
 
 
-class ModuleDocFragment(object):
+
+class ModuleDocFragment:
     # Common parameters for Proxmox VE modules
     DOCUMENTATION = r"""
 options:
--- a/plugins/doc_fragments/proxmox.py
+++ b/plugins/doc_fragments/proxmox.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 # Copyright (c) Ansible project
 # GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
 # SPDX-License-Identifier: GPL-3.0-or-later
@@ -56,7 +55,7 @@ options:
     default: 5
   validate_certs:
     description:
-      - Validate the TLS certificates used for the connection to the Proxmox VE API.  
+      - Validate the TLS certificates used for the connection to the Proxmox VE API.
       - Currently defaults to V(false) and changes default to V(true) with community.proxmox 2.0.0.
       - Uses the E(PROXMOX_VALIDATE_CERTS) environment variable if not specified.
     type: bool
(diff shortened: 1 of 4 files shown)


Thulium-Drake:
`ca_path` to me sounds sensible :-)

WRT changing the default, we can start by making the current default deprecated and have some time for it to brew before flipping the switch.

I, however, do not know to mark a value as deprecated :sweat_smile: maybe @felixfontein can help us out there :innocent: ?

plugins/doc_fragments/proxmox.py
```diff
@@ -44,13 +44,18 @@ class ModuleDocFragment(object):
       - Specify the token secret.
       - Uses the E(PROXMOX_TOKEN_SECRET) environment variable if not specified.
     type: str
+  ca_path:
+    description:
+      - Path to a local certificate, which will be used to verify tls connections.
```
felixfontein: https://docs.ansible.com/projects/ansible/devel/dev_guide/style_guide/spelling_word_choice.html#acronyms
```suggestion
      - Path to a local certificate, which will be used to verify TLS connections.
```

plugins/doc_fragments/proxmox.py
```diff
@@ -44,13 +44,18 @@ class ModuleDocFragment(object):
...
+      - Ignored if O(validate_certs=false).
+    type: str
```
felixfontein: Since this is a path:
```suggestion
    type: path
```

felixfontein:
> WRT changing the default, we can start by making the current default deprecated and have some time for it to brew before flipping the switch.
> 
> I, however, do not know to mark a value as deprecated 😅 maybe @felixfontein can help us out there 😇 ?

There's no specific mechanism for this, the usual procedure is:

1. Remove the default, and add code that uses the previous default if the value is `None`, and adjust the documentation to say in `description` what the current default is.
2. In the code that handles `None`, add a `module.deprecate()` call that informs the user that the current default is deprecated and will change in community.proxmox X.0.0. Tell the user to explicitly set the value to avoid deprecation warnings.

plugins/module_utils/proxmox.py
```diff
@@ -153,15 +148,20 @@ def _connect(self):
         api_password = self.module.params['api_password']
         api_token_id = self.module.params['api_token_id']
         api_token_secret = self.module.params['api_token_secret']
+        if self.module.params['validate_certs'] is None:
+            self.module.deprecate("The connection setting `validate_certs` was not provided and "
```
Thulium-Drake: Shouldn't this be the other way around? It defaults to `false` (not validating) and it will start defaulting to `true` (do validate) :-)

IamLunchbox: Yeah my bad!

felixfontein:
Hmm, the current ruff config exposes a bug in antsibull-nox, since it now has trouble parsing the `ruff check` output. (This is basically caused by nox interleaving stdout and stderr when capturing output. The decision to do that is once again biting me...)

IamLunchbox:
Well it is not only you being bitten by this PR, since I have to fix all the new requirements for proxmox.py now :laughing: 

Regarding the bug: Shouldn't community.general have the same problem? :thinking:

felixfontein:
community.general doesn't have this problem since it does not select the `DOC` rules (and thus does not trigger the "`warning: Selection `DOC` has no effect because preview is not enabled.`" warning).

felixfontein:
Also there is a second bug: if change detection is enabled, and a linter config file is changed, all files affected by that config have to be checked (and not just the changed ones).

IamLunchbox:
> community.general doesn't have this problem since it does not select the `DOC` rules (and thus does not trigger the "`warning: Selection `DOC` has no effect because preview is not enabled.`" warning).

Luckily, DOC is in preview and is disabled again, because I did not want to enable those.

IamLunchbox:
CI should pass as soon as #274 is merged. 

I formatted the file within this PR, I know this is ugly and could be a separate PR. But since I changed the core proxmox.py I wanted all tests to pass.

felixfontein:
You could rebase the PR to see whether it actually passes.

IamLunchbox:
Well, yes. But I did not think about that until now. 😞

Thulium-Drake:
You'll need to do the rebase manually, as there seems to be some conflict with the `main` branch now :-) I can't do it for you using the fancy buttons GH makes ;-)

felixfontein:
> community.general doesn't have this problem since it does not select the `DOC` rules (and thus does not trigger the "`warning: Selection `DOC` has no effect because preview is not enabled.`" warning).

I've implemented a workaround, namely passing `--quiet` which stops ruff from printing such warnings. It's currently only in the `main` branch, but at least for CI this should fix the issue ;)

IamLunchbox:
I think we use `main` implicitly, I'll open a MR and add `--quiet`.

Thanks for implementing a workaround so swiftly :)

Thulium-Drake:
@IamLunchbox Can you rebase this PR on your system? :-) I can't rebase it for you..

IamLunchbox:
i'd wait until #279  is finished, merge and then apply formatting, if it needs to.

IamLunchbox:
I fixed the linting errors in proxmox.py as well. But i had to edit `proxmox_storage_info` as well, since it inherited from the proxmox base class and changed a function - so I streamlined the two function names.

plugins/module_utils/proxmox.py
```diff
@@ -129,6 +161,21 @@ def _connect(self):
...
         api_token_id = self.module.params["api_token_id"]
         api_token_secret = self.module.params["api_token_secret"]
+        if self.module.params["validate_certs"] is None:
+            self.module.deprecate(
+                "The connection setting `validate_certs` was not provided and "
+                "defaults to `false`. This default will change to `true` in"
+                "in community.proxmox 2.0.0.",
+                version="2.0.0",
+                collection_name="community.proxmox",
+            )
```
felixfontein, line 171: The deprecation should also be mentioned in the changelog fragment (`deprecated_features:` section).

Thulium-Drake:
I made a task to remind us that we need to do this once 2.0.0 lands (which could be in a while)

felixfontein:
@Thulium-Drake once you change the version in galaxy.yml to 2.0.0, the sanity tests will remind you (by failing) that there's a `AnsibleModule.deprecate` call with `version=2.0.0` :)

Note that this check only works with certain newer versions of ansible-core if you have type hints or the type can be inferred. (Like if you change the constructor to `def __init__(self, module: AnsibleModule):`.) (This changed in ansible-core 2.19; ref: https://github.com/ansible/ansible/issues/85150.)

RussellAult:
As it turns out, @IamLunchbox has (unknowingly, it seems) resolved a feature request that I submitted as issue #62 over two years ago; thank you, @IamLunchbox !

PR #257, opened 2026-01-19
Improve inventory speed

aadnehovda:
##### SUMMARY
Using the inventory plugin is very slow due to multiple looping API calls, especially over WAN links. This PR addresses this in four ways:

1. Change the node list API call from `/api2/json/nodes` to `/api2/json/cluster/status`, since the latter already includes the management IP in the payload.
2. Fetch QEMU and LXC inventory records from `/api2/json/cluster/resources?type=vm` instead of querying every online node's QEMU and LXC endpoints.
3. Add `exclude_vms` so users can skip QEMU/LXC discovery entirely when they only need Proxmox nodes.
4. Add `facts_concurrency` so users can opt into bounded concurrent guest fact collection when `want_facts: true`.

Fetching the inventory for nodes and VMs is now done in one or two single requests, which cuts the wait dramatically. When you need VM facts, it fans out and fetches them in parallel if you want, which also add large time savings. These fixes make it possible to use the inventory plugin against some of my big remote clusters, whereas before it could take 10-20+ minutes or simple never complete.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
Inventory

##### ADDITIONAL INFORMATION

Sample response from `/api2/json/cluster/status`
```json
{
    "data": [
        {
            "type": "cluster",
            "quorate": 1,
            "name": "svg001pve",
            "version": 24,
            "id": "cluster",
            "nodes": 3
        },
        {
            "level": "",
            "id": "node/svg025pve",
            "online": 1,
            "type": "node",
            "local": 1,
            "nodeid": 3,
            "ip": "192.168.7.75",
            "name": "svg025pve"
        },
        {
            "id": "node/svg013pve",
            "online": 1,
            "level": "",
            "name": "svg013pve",
            "ip": "192.168.7.63",
            "nodeid": 2,
            "type": "node",
            "local": 0
        },
        {
            "level": "",
            "online": 1,
            "id": "node/svg011pve",
            "local": 0,
            "type": "node",
            "nodeid": 1,
            "ip": "192.168.7.61",
            "name": "svg011pve"
        }
    ]
}
```

The inventory plugin now uses `/api2/json/cluster/resources?type=vm` for the initial QEMU/LXC records. This keeps the existing host variables and groups, while avoiding separate `/nodes/{node}/qemu` and `/nodes/{node}/lxc` list calls for each online node. Full guest facts are still fetched separately when `want_facts` or `want_post_filter_facts` requires them.

Sample response from `/api2/json/cluster/resources?type=vm`
```json
{
    "data": [
        {
            "id": "qemu/101",
            "name": "app01",
            "node": "pve01",
            "pool": "production",
            "status": "running",
            "template": 0,
            "type": "qemu",
            "vmid": 101
        },
        {
            "id": "lxc/201",
            "name": "web01",
            "node": "pve02",
            "pool": "production",
            "status": "running",
            "tags": "linux;web",
            "template": 0,
            "type": "lxc",
            "vmid": 201
        }
    ]
}
```

New option to exclude VM discovery:
```yaml
plugin: community.proxmox.proxmox
url: https://pvehost:8006
want_proxmox_nodes_ansible_host: true
exclude_nodes: false
exclude_vms: true
```

New option to gather guest facts concurrently:
```yaml
plugin: community.proxmox.proxmox
url: https://pvehost:8006
want_facts: true
facts_concurrency: 8
```

`facts_concurrency` defaults to `1`, which preserves the previous serial behavior. Higher values can reduce runtime over slow links, but increase load on the Proxmox API.

Performance highlights from live inventory runs against remote cluster (~100ms ping) with 24 Nodes, ~450 VMs (qemu/lxc/templates):

| Configuration | Before | After |
| --- | ---: | ---: |
| `want_facts: false` | `1m:46s` | `3.4s` |
| `want_facts: false`, `exclude_vms: true` | `1m:44s` | `0.9s` |
| `want_facts: true`, `facts_concurrency: 8` | `13m15s` | `1m52s` |

##### TODO
Consider migrating to `proxmoxer`.

### DIFF
--- a/plugins/inventory/proxmox.py
+++ b/plugins/inventory/proxmox.py
@@ -91,6 +91,14 @@ DOCUMENTATION = """
             but its actual state will be paused. See O(qemu_extended_statuses) for how to retrieve the real status.
         default: false
         type: bool
+      facts_concurrency:
+        description:
+          - Number of concurrent workers to use when gathering LXC/QEMU configuration facts.
+          - Only applies when O(want_facts) is set to V(true).
+          - Higher values can reduce inventory runtime over slow links, but will increase load on the Proxmox API.
+          - Set to V(1) to gather facts serially.
+        default: 1
+        type: int
       want_post_filter_facts:
         description:
         - Whether to collect facts after host filtering (in contrast to pull all facts of all hosts before filtering as with O(want_facts) set to V(true)).
@@ -113,6 +121,10 @@ DOCUMENTATION = """
         default: false
       exclude_nodes:
         description: Exclude proxmox nodes and the nodes-group from the inventory output.
+        type: bool
+        default: false
+      exclude_vms:
+        description: Exclude LXC containers and QEMU virtual machines from the inventory output.
         type: bool
         default: false
       filters:
@@ -209,10 +221,11 @@ want_proxmox_nodes_ansible_host: true
 
 """
 
-import itertools
 import re
 from collections.abc import MutableMapping
+from concurrent.futures import ThreadPoolExecutor, as_completed
 from sys import version as python_version
+from threading import Lock, local
 from urllib.parse import urlencode
 
 from ansible.errors import AnsibleError
@@ -247,9 +260,11 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
         # from config
         self.proxmox_url = None
 
-        self.session = None
+        self._thread_local = local()
+        self._results_lock = Lock()
         self.cache_key = None
         self.use_cache = None
+        self.facts_concurrency = 1
 
     def verify_file(self, path):
         valid = False
@@ -261,13 +276,15 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
         return valid
 
     def _get_session(self):
-        if not self.session:
-            self.session = requests.session()
-            self.session.headers.update(
+        session = getattr(self._thread_local, "session", None)
+        if session is None:
+            session = requests.session()
+            session.headers.update(
                 {"User-Agent": f"ansible {ansible_version} Python {python_version.split(' ', 1)[0]}"}
             )
-            self.session.verify = self.get_option("validate_certs")
-        return self.session
+            session.verify = self.get_option("validate_certs")
+            self._thread_local.session = session
+        return session
 
     def _get_auth(self):
         if self.proxmox_password:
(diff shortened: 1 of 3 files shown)


IamLunchbox:
Hi, thanks for the pr. Remember to create a changelog fragment :)

aadnehovda:
> Hi, thanks for the pr. Remember to create a changelog fragment :)

Rebased and added changelog. Due to the api usage change there's now a failing test which I assume must be adapted before this will get approved.

aadnehovda:
Found it hard a little tricky to run the tests, but I believe they should work now.

IamLunchbox:
@aadnehovda Could you pull in the current main branch into your PR and apply the conflict resolution? We recently enabled linting and formatting and ci and so all files changed.

Thulium-Drake:
I have started the workflows on the PR, so that should provide you with some more guidance :-)

Do note, your PR might have touched files that contain a lot of linting errors, you don't need to fix any of the errors that isn't in your diff. We'll pick those up later.

In a nutshell, if the unit and 'normal' sanity tests pass, it's good to go and we'll look at it, provide feedback and/or merge it when it's done :-)

aadnehovda:
> I have started the workflows on the PR, so that should provide you with some more guidance :-)

Thanks, think I've found the final issue with the failing test now. Please run it again.

Thulium-Drake:
If you don't already know, you can see the results of the CI jobs here: https://github.com/ansible-collections/community.proxmox/pull/257/checks

Or in the box at the bottom of the thread, you can just click the failing job and it will bring you right over to the output from the tests :-)

IamLunchbox (changes requested):
Thank you for your effort, I did not dive too deep into this module until now. But it seems the overall code could really need some improvement, directly working with requests does not really make sense, when we have proxmoxer certainly available - all other modules use it.

Regarding your PR: LGTM, but could  you add unit tests for your two new options?

And nice idea to query cluster.status - it should really save so much time on multi cluster setups. We could actually speed up discovering qemu and lxc further, if we'd use cluster.resources instead of blindly querying each node.qemu/node.lxc.

aadnehovda:
> Thank you for your effort, I did not dive too deep into this module until now. But it seems the overall code could really need some improvement, directly working with requests does not really make sense, when we have proxmoxer certainly available - all other modules use it.

I agree, a refactor would be good, however one limitation with proxmoxer is that afaict it does not support async, https://github.com/proxmoxer/proxmoxer/issues/209 .

> Regarding your PR: LGTM, but could you add unit tests for your two new options?

Yep, will do.

> And nice idea to query cluster.status - it should really save so much time on multi cluster setups. We could actually speed up discovering qemu and lxc further, if we'd use cluster.resources instead of blindly querying each node.qemu/node.lxc.

I thought the same, it's a good match when facts are not wanted/needed. So initial discovery via cluster.resources -> then async fetch node/qemu/lxc facts (if wanted, and optionally filtered) would be the absolute quickest. Until perhaps Proxmox adds a GraphQL wrapper around the REST API sometime in the future and we could get everything in a single roundtrip.

IamLunchbox:
I doubt that graphql will happen though. But one can hope 🙏

We could speed up requests by using threading though. I doubt in the inventory generation we'd encounter race conditions. And while it is not async, it is still faster then sequentially polling each endpoint. 

My dynamic inventory needs 10 seconds for 20 VMs, I don't want to know what larger corporations need to wait for....

aadnehovda:
Added the other perf updates too. Inventory now uses the optimized endpoints, you can skip vm inventory and facts can be gathered in parallel.

aadnehovda:
> I have started the workflows on the PR, so that should provide you with some more guidance :-)

Again please, @Thulium-Drake 😊

IamLunchbox:
@aadnehovda Awsome, that you further worked on that! I think you may have to format your code and then you'll get to see the linting errors through ci as well. We recently wrote a better guide in [CONTRIBUTING.md](https://github.com/ansible-collections/community.proxmox/blob/main/CONTRIBUTING.md), if you need more guidance on which tools and commands you need to do that locally. :)

plugins/inventory/proxmox.py
```diff
@@ -91,6 +91,20 @@
...
         type: bool
+      facts_concurrency:
+        description:
+          - Number of concurrent workers to use when gathering LXC/QEMU configuration facts.
+          - Only applies when O(want_facts) is set to V(true).
+          - Higher values can reduce inventory runtime over slow links, but will increase load on the Proxmox API.
+          - Set to V(1) to gather facts serially.
+        default: 1
+        type: int
+      request_timeout:
```
IamLunchbox: All other modules have the options `api_timeout` through  `module_utils/proxmox.py`, maybe we could add this as an alias here to streamline the api connection options?

plugins/inventory/proxmox.py
```diff
@@ -247,9 +266,12 @@ def __init__(self):
         # from config
         self.proxmox_url = None
 
-        self.session = None
+        self._thread_local = local()
+        self._results_lock = Lock()
         self.cache_key = None
         self.use_cache = None
+        self.facts_concurrency = 1
+        self.request_timeout = 30
```
IamLunchbox: The default from the other modules is 5, would you mind to use that value as well? Or do you think this is too low, since the calls usually take longer for enumerating cluster data?

IamLunchbox (approved):
Minor questions regarding compatibility to the other modules. Otherwise, i am fine to merge :)

And you changes really improved the inventory speed - even though my testcluster is not too large - so it might not be a too good test tbh.

plugins/inventory/proxmox.py
aadnehovda: Since it's a new option let's just rename it

plugins/inventory/proxmox.py
aadnehovda: 5 is okay in my testing

Thulium-Drake:
CI's happy, I'm happy ;-)

Thanks for making this! :rocket:

PR #261, opened 2026-01-22
inventory: add want_post_filter_facts

sdubrul:
##### SUMMARY
The `want_post_filter_facts` option makes it possible to delay fact gathering until host filtering has completed. 
The idea is to speed up the inventory creation by avoiding `want_facts` of all guests, which seems to be the most time‑consuming part.
Filtering still works as expected, but is limited to a reduced set of early facts such as name, vmid, status, node and tags.

##### ISSUE TYPE
Feature Pull Request

##### COMPONENT NAME
inventory plugin

##### ADDITIONAL INFORMATION
```paste below
# create a proxmox.yml in your inventory with the additional want_post_filter_facts: True
plugin: community.proxmox.proxmox
strict: True
url: "https://example.com:8006"
validate_certs: False
user: "root@pam"
password: "xxxx"
want_facts: False
filters:
- "'production-' in (proxmox_name | default(''))"
want_post_filter_facts: True

# run ansible-inventory as before
ansible-inventory -i inventories/prox --list
```

### DIFF
--- a/changelogs/fragments/261-want_post_filter_facts.yml
+++ b/changelogs/fragments/261-want_post_filter_facts.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - inventory plugin - add want_post_filtering_facts to delay fact gathering until filtering has completed (https://github.com/ansible-collections/community.proxmox/pull/261).


aadnehovda:
Does it make sense to merge this change with #257 since they both seem to address related issues?

aadnehovda:
Anwering my own question after a closer look... the changes don't seem to overlap so probably no :-)

Thulium-Drake:
Thanks for this! :rocket:

Issue #234, opened 2025-12-08
Add support for waiting until snapshot creation is completed


Tealk:
##### SUMMARY
Add support for waiting until Proxmox snapshot creation is completed before proceeding to the next task.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
proxmox_snap

##### ADDITIONAL INFORMATION
Currently, when creating a snapshot in Proxmox, the Ansible task completes and proceeds to the next task before the snapshot operation has actually finished on the Proxmox server.

**Proposed Solution:**
Add a parameter (e.g., `wait: yes` or `wait_for_completion: yes`) that makes the task poll the Proxmox API until the snapshot operation is complete before marking the task as finished.

PR #264, opened 2026-01-25
Validate unbind auth at module init

PendaGTP:
##### SUMMARY

Run unbind validation (`root@pam` + `api_password`) during module setup to fails early.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

`proxmox_snap`

##### ADDITIONAL INFORMATION

This is part of https://github.com/ansible-collections/community.proxmox/issues/234.

To keep the diff small and reviewable, I will split into smaller PRs, further consolidation will follow.

### DIFF
--- a/plugins/modules/proxmox_snap.py
+++ b/plugins/modules/proxmox_snap.py
@@ -149,7 +149,7 @@ from ansible_collections.community.proxmox.plugins.module_utils.proxmox import (
 )
 
 
-def get_proxmox_args():
+def get_ansible_module_args():
     return dict(
         hostname=dict(type="str"),
         vmid=dict(type="str"),
@@ -166,7 +166,7 @@ def get_proxmox_args():
 
 def get_ansible_module():
     module_args = proxmox_auth_argument_spec()
-    module_args.update(get_proxmox_args())
+    module_args.update(get_ansible_module_args())
 
     return AnsibleModule(
         argument_spec=module_args,
@@ -183,8 +183,7 @@ def validate_params(params):
     # API tokens are not supported for mountpoint operations, even for root@pam.
     # Without proper authentication, the operation would fail after mountpoints are removed,
     # leaving the container in a misconfigured state without its mountpoints.
-    unbind = params.get("unbind", False)
-    if unbind is True:
+    if params.get("unbind"):
         api_user = params.get("api_user")
         api_password = params.get("api_password")
         if api_user != "root@pam" or not api_password:


IamLunchbox:
How does that relate to #234? 🤔

PendaGTP:
I'm currently working on #234.
I have a poc working and to avoid creating a PR with lot of diff I prefer split it into smaller PRs.

PendaGTP:
FYI now in draft related to:
- #272
- https://github.com/ansible-collections/community.proxmox/pull/274#issuecomment-3812284154

PendaGTP:
@IamLunchbox What should I do: lint/format the whole file `proxmox_snap.py`? Or only changed lines on my PR?

Related to: https://github.com/ansible-collections/community.proxmox/pull/274#issuecomment-3812284154

> Edit: I also think it's better for the repository maintainers to lint/format all existing code, and for PRs, rebasing will suffice. It will be much easier to review PRs if they don't include all the style changes due to changes in the linter/formatter configuration.


Same question for my other PRs.

PendaGTP:
> @IamLunchbox What should I do: lint/format the whole file `proxmox_snap.py`? Or only changed lines on my PR?
> 
> Related to: [#274 (comment)](https://github.com/ansible-collections/community.proxmox/pull/274#issuecomment-3812284154)
> 
> > Edit: I also think it's better for the repository maintainers to lint/format all existing code, and for PRs, rebasing will suffice. It will be much easier to review PRs if they don't include all the style changes due to changes in the linter/formatter configuration.
> 
> Same question for my other PRs.

I just added lint/format for my changes.

Let me know if need anything, thanks u for your time!

IamLunchbox:
Yes, its enough to lint your part. :)

The rest can and should happen separately - in due time. Until then we have to live with failing tests.

PendaGTP:
Thanks, so this one is now ready to review. Let me know if anything wrong.

I will update others!

PendaGTP:
FYI: rebased, formated, linted

Ready!

plugins/modules/proxmox_snap.py
```diff
@@ -148,6 +148,48 @@
...
 
+def get_proxmox_args():
+    return dict(
+        hostname=dict(),
+        vmid=dict(),
+        state=dict(default="present", choices=["present", "absent", "rollback"]),
+        force=dict(type="bool", default=False),
+        unbind=dict(type="bool", default=False),
+        vmstate=dict(type="bool", default=False),
+        description=dict(),
```
IamLunchbox: Could you add the types here?

plugins/modules/proxmox_snap.py
```diff
@@ -148,6 +148,48 @@
...
+    # Mountpoint configuration operations require root@pam with password authentication.
+    # API tokens are not supported for mountpoint operations, even for root@pam.
+    # Without proper authentication, the operation would fail after mountpoints are removed,
+    # leaving the container in a misconfigured state without its mountpoints.
+    unbind = params.get("unbind", False)
+    if unbind is True:
+        api_user = params.get("api_user")
+        api_password = params.get("api_password")
+        if api_user != "root@pam" or not api_password:
+            raise Exception("Parameter 'unbind=True' requires 'api_user' and 'api_password' options.")
```
IamLunchbox: Is there a specific point why you raise a broad `Exception`?  Maybe [AnsibleOptionsError](https://github.com/ansible/ansible/blob/be09a21446e34ccce341b49e69cc674c2d9049ac/lib/ansible/errors/__init__.py#L186) is more specific and you don't need to use `except Exception` later on - even though ansible probably does not mind if keyboard interrupts are caught.

plugins/modules/proxmox_snap.py
PendaGTP: The default type is str, what's missing?

plugins/modules/proxmox_snap.py
PendaGTP: The idea was to keep `validate_params()` independent from Ansible-specific error handling and avoid duplicating `module.fail_json()` calls as we add more validation rules.

That said, I’m open to switching to a more specific exception type, I wasn't aware of the existence of `AnsibleOptionsError`.

It would be really nice to align on a common validation pattern for the collection, and apply it to the other modules.

plugins/modules/proxmox_snap.py
IamLunchbox: Ansible validates the input against the values passed as `type` .

So only `hostname=dict(type="str")` ensures, that actually a string is in module.params["hostname"] and invalid inputs get handled by Ansible with the appropiate error and failure message.

plugins/modules/proxmox_snap.py
IamLunchbox: I think `AnsibleError` is sensible when we work within the module and don't want to fail immediately. 

More often then not you'll see `except Exception` in community.proxmox, because some following code calls proxmoxer and the range of different errors, which could be raised there, is not known. 


....and much code is copied from other modules so if one developer did this in the past, it populated into all of the plugins copies. 

I think the [developer docs](https://docs.ansible.com/projects/ansible/latest/dev_guide/developing_modules_best_practices.html#handling-exceptions-bugs-gracefully) have some room of interpretation. Yes, avoid stacktraces, but also don't catch Exceptions to broadly. 

In your case I'd say AnsibleError or a subclass is very fitting. For the existing code which wraps proxmoxer: I am not sure, I can imagine why it is so broadly, but it is not too sensible. On the other hand this definetely avoids stack traces. 😅

plugins/modules/proxmox_snap.py
IamLunchbox: Thanks for fixing this, even though this is not your original code! :)

A little nitpick though: vmid is an int  😇

plugins/modules/proxmox_snap.py
PendaGTP: Thanks that makes sense!

PendaGTP:
@IamLunchbox now should be good, sorry localy I was missing to run sanity checks...

plugins/modules/proxmox_snap.py
```diff
@@ -141,13 +141,56 @@
...
+    # Mountpoint configuration operations require root@pam with password authentication.
+    # API tokens are not supported for mountpoint operations, even for root@pam.
+    # Without proper authentication, the operation would fail after mountpoints are removed,
+    # leaving the container in a misconfigured state without its mountpoints.
+    unbind = params.get("unbind", False)
+    if unbind is True:
+        api_user = params.get("api_user")
+        api_password = params.get("api_password")
+        if api_user != "root@pam" or not api_password:
+            raise AnsibleValidationError("Parameter 'unbind=True' requires 'api_user' and 'api_password' options.")
```
IamLunchbox:
```suggestion
    if params.get("unbind"):
        api_user = params.get("api_user")
        api_password = params.get("api_password")
        if api_user != "root@pam" or not api_password:
            raise AnsibleValidationError("Parameter 'unbind=True' requires 'api_user' and 'api_password' options.")
```

Just generally speaking, the "more" pythonic way is just `if variable:`, if your variable can be evaluated with a simple `bool()`. You need `is` only in specific circumstances (usually object identity), which are not met here.

In this case `dict.get()` returns `None` when the key does not exists, which in turn evaluates to `False`.

plugins/modules/proxmox_snap.py
```diff
@@ -141,13 +141,56 @@
...
 from ansible.module_utils.common.text.converters import to_native
+from ansible.module_utils.errors import AnsibleValidationError
 
 from ansible_collections.community.proxmox.plugins.module_utils.proxmox import (
     ProxmoxAnsible,
     proxmox_auth_argument_spec,
 )
 
 
+def get_proxmox_args():
```
IamLunchbox:
```suggestion
def get_snapshot_args():
```
Or any other name you prefer. But proxmox could be mixed up with the general class or auth IMHO.

IamLunchbox (approved):
Nitpicking, otherwise LGTM

Thulium-Drake:
Looks good! If you can look into the nitpickings, I'll happily merge it! :-)

plugins/modules/proxmox_snap.py
PendaGTP: :heavy_check_mark:

plugins/modules/proxmox_snap.py
PendaGTP: :heavy_check_mark:

PendaGTP:
@Thulium-Drake Done! Let me know if any thing need changes.

Thulium-Drake:
Thanks for this work! :-) :rocket:

PR #274, opened 2026-01-28
Update ruff.toml with new linting rules

IamLunchbox:
Removed commented-out rules and added new linting rules for docstring formatting. More specifically, removed mutually exclusive rules.

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -8,7 +8,7 @@ include = ["tests/**/*.py", "plugins/**/*.py"]
 [lint]
 # https://docs.astral.sh/ruff/rules/
 
-select = ["A", "B", "E", "D", "F", "FA", "FLY", "UP", "SIM", "I", "RUF022", "PL"]
+select = ["A", "B", "D", "DOC", "E", "F", "FA", "FLY", "UP", "SIM", "I", "RUF022", "PL"]
 ignore = [
     # Better keep ignored (for now)
     "F811",  # Redefinition of unused `xxx` (happens a lot for fixtures in unit tests)


IamLunchbox:
@Thulium-Drake as you see, I changed ruff.toml and all code was linted as well.

There is this configuration, https://docs.ansible.com/projects/antsibull-nox/change-detection/, maybe that would help and buy time until all modules are properly onboarded to linting. 

I just copied partials from community.general, Ill copy their vcs settings as well

PendaGTP:
@IamLunchbox @Thulium-Drake

Not directly related to this PR but: I really think keeping documentation rules is a bad idea, as it will only add unnecessary function descriptions everywhere.

Edit: may only be necessary for `plugins/module_utils` and ONLY add func doc when necessary (e.g. it is not easy to understand what the func does without reading the entire func code).

Edit: I also think it's better for the repository maintainers to lint/format all existing code, and for PRs, rebasing will suffice. It will be much easier to review PRs if they don't include all the style changes due to changes in the linter/formatter configuration.

These are just my thoughts; it's up to you to decide how you want to proceed. I will adapt and update my open PRs once the style changes are stable.

IamLunchbox:
This is where we differ from community.general - I added this setting. But I will check on ruffs docs if we can relax it, e.g. to make sure only large functions need docstrings. On the other hand I really value docstrings, because we have changing contributors and it makes understanding the code easier. 

And in times of LLMs generating these is way less work then it was before. 

But first thank you for feedback, I appreciate it! 

Regarding the linting of the whole codebase, I am checking right now what needs to be configured to make this possible - I agree, that the current state is not optimal.

IamLunchbox:
Hey, @felixfontein - i am sorry to bother you.

But I tried to enable change detection with antsibull-nox, since I plan to fix the linting of existing modules step by step.

With the above changes, `antsibull-nox show-changes` and `uv run noxfile.py -e lint` report and lint the correct files.

But the ci (importing your gh action <3 ) fails when I provide it with `change-detection: true`:

```
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.14.2/x64/lib/python3.14/site-packages/nox/sessions.py", line 1119, in execute
    self.func(session)
    ~~~~~~~~~^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.14.2/x64/lib/python3.14/site-packages/nox/_decorators.py", line 99, in __call__
    return self.func(*args, **kwargs)
           ~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.14.2/x64/lib/python3.14/site-packages/antsibull_nox/sessions/extra_checks.py", line 159, in extra_checks
    execute_no_unwanted_files(session)
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.14.2/x64/lib/python3.14/site-packages/antsibull_nox/sessions/extra_checks.py", line 97, in execute_no_unwanted_files
    run_bare_script(
    ~~~~~~~~~~~~~~~^
        session,
        ^^^^^^^^
    ...<15 lines>...
        process_messages=True,
        ^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/opt/hostedtoolcache/Python/3.14.2/x64/lib/python3.14/site-packages/antsibull_nox/sessions/utils/scripts.py", line 49, in run_bare_script
    files = filter_files_cd(files)
  File "/opt/hostedtoolcache/Python/3.14.2/x64/lib/python3.14/site-packages/antsibull_nox/sessions/utils/paths.py", line 103, in filter_files_cd
    changed_files = get_changes(relative_to=Path.cwd())
  File "/opt/hostedtoolcache/Python/3.14.2/x64/lib/python3.14/site-packages/antsibull_nox/cd.py", line 143, in get_changes
    changes = cd_config.provider.get_changes_compared_to(
        repo=cd_config.repo, branch=cd_config.base_branch
    )
  File "/opt/hostedtoolcache/Python/3.14.2/x64/lib/python3.14/site-packages/antsibull_nox/vcs/git.py", line 178, in get_changes_compared_to
    raise ValueError(
        f"Cannot determine fork point from current branch/HEAD to {branch}!"
    )
ValueError: Cannot determine fork point from current branch/HEAD to main!
```

I could not find any more configuration possibilities for this feature and it works locally - so do you have an idea why it fails in this instance?

felixfontein:
@IamLunchbox the problem is that you're using the action without setting the Git repo up for change detection. Better use the shared workflow https://github.com/ansible-community/antsibull-nox/blob/main/.github/workflows/reusable-nox-run.yml which does this for you. After setting up some variables, it does:
```
      - name: Check out collection
        uses: actions/checkout@v6
        with:
          persist-credentials: false
          fetch-depth: ${{ steps.determine-parameters.outputs.change-detection == 'true' && '0' || '1' }}
      - name: Create branches for change detection
        if: >-
          steps.determine-parameters.outputs.change-detection == 'true'
        env:
          GHA_PR_NUMBER: ${{ github.event.pull_request.number }}
          GHA_BASE_REF: ${{ github.event.pull_request.base.ref }}
        run: |
          # Create a branch for the current HEAD, which happens to be a merge commit
          git checkout -b "pull-request-${GHA_PR_NUMBER}"

          # Name the target branch
          git branch "${GHA_BASE_REF}" --track "origin/${GHA_BASE_REF}"

          # Show branch information
          git branch -vv
        shell: bash
```
This would simplify your workflow to basically this: https://github.com/ansible-collections/community.docker/blob/main/.github/workflows/nox.yml#L19-L23

(According to the docs it might be that `fetch-depth: 0` for the checkout action would be sufficient: https://github.com/actions/checkout?tab=readme-ov-file#fetch-all-history-for-all-tags-and-branches, you could also test that first)

antsibull-nox.toml
```diff
@@ -8,13 +8,26 @@
...
 ruff_check_config = "ruff.toml"
 run_ruff_format = true
 ruff_format_config = "ruff.toml"
+run_ruff_autofix = true
+ruff_autofix_config = "ruff.toml"
+ruff_autofix_select = [
+    "I",
+    "RUF022",
+]
+run_ruff_check = true
```
felixfontein: You have `run_ruff_check = true` twice in this section.

IamLunchbox:
> (According to the docs it might be that fetch-depth: 0 for the checkout action would be sufficient: https://github.com/actions/checkout?tab=readme-ov-file#fetch-all-history-for-all-tags-and-branches, you could also test that first)

Is that not contradictory to "just" using the reusable nox workflows? 

Anyways, thank you for providing the shared workflow! Regarding the license in ruff.toml, I kept your license because I more or less copied it. Am I required to remove or append to it when I change the file?

PendaGTP:
> This is where we differ from community.general - I added this setting. But I will check on ruffs docs if we can relax it, e.g. to make sure only large functions need docstrings. On the other hand I really value docstrings, because we have changing contributors and it makes understanding the code easier.
> 
> And in times of LLMs generating these is way less work then it was before.
> 
> But first thank you for feedback, I appreciate it!
> 
> Regarding the linting of the whole codebase, I am checking right now what needs to be configured to make this possible - I agree, that the current state is not optimal.

@IamLunchbox

I still don't think enforcing doc rules is a good idea:

- Most modules are small and mainly use internal module functions
- Clear function names should be sufficient, if a function needs a docstring to explain what it does, that may indicate it should be refactored/review
- Mandatory func doc everywhere will add noise rather than improving readability
- It also encourages "vibe-coding" style contributions, which I’m personally not a fan

I'm more fine with:

- Docstrings for shared/utils functions across modules
- Inline comments for non-obvious logic or design decisions
- Adding a docstring when a reviewer explicitly ask because it seems needed

Just my personal opinion, no offense intended, and thank you for being open to feedback

IamLunchbox:
Hey @PendaGTP,
I just updated the rules, but your feedback came in parallel. First of all, I removed the need for docstrings in test files.

What I am ultimately going for is improved editor support to support development. Since type annotations may vary from developer to developer, docstrings can offer great editor support without being dependent on the used python version (e.g. using `typing` or native builtin types). 

The main beneficial point of requiring docstrings is being transparent about their creation and giving an early indiciation in CI, that the documentation is required - and saving time of maintainers reading code and asking for changes.

I, personally, don't believe that asking for docstrings would  push  users to use LLMs and vibe code the docs - you could say the same about pylint or flake8 - why shouldn't a developer user a LLM at that moment to fix these errors?

And while it is true, that a well written function, by using understandable function and variable names, possibly gets rid of the need to write docstrings, it is not always the case for all contributors. Always asking for comments and docstrings of a case by case basis also leads to friction and mixed code quality.

I, myself, am hesitant which way to go. Only requiring docstrings for `module_utils/proxmox[_sdn].py` could be a middle ground for now and I am open for that.

Maybe @russoz, @Thulium-Drake and @felixfontein could share their experience as maintainers of community.general and community.proxmox? :)

felixfontein:
I guess you also have to consider that a badly written docstring is less helpful than no docstring (since you try understanding it, or misunderstand it, instead of simply looking at the code).

I guess in the end it doesn't really matter, in the end you have to carefully review submissions if you want good quality code in the collection.

In any case, the middle ground you're proposing sounds good to me, it's a good compromise between annoying contributors (do I really have to write a docstring that is longer than the actual content of the function?!) and making sure that at least shared code has documentation. (You still have to ensure it is good enough during review... `"Blah."` is also a valid docstring accepted by most linters ;) )

felixfontein:
> > (According to the docs it might be that fetch-depth: 0 for the checkout action would be sufficient: https://github.com/actions/checkout?tab=readme-ov-file#fetch-all-history-for-all-tags-and-branches, you could also test that first)
> 
> Is that not contradictory to "just" using the reusable nox workflows?

It depends on your preference. Maybe there was a special reason why the workflow used the shared workflow for the ansible-test matrix, but the action for the single run. Maybe it was also just a leftover from before the shared workflow for a single run existed. (I think I only added it when adding change detection, since before that it wasn't really necessary.)

Also I was curious whether `fetch-depth: 0` without creating the branch in question would work. (IIRC it didn't somewhen in the past, but that might have changed inbetween.) ;)

> Anyways, thank you for providing the shared workflow! Regarding the license in ruff.toml, I kept your license because I more or less copied it. Am I required to remove or append to it when I change the file?

russoz:
@IamLunchbox I reckon that there are some objective statements we can make in this topic, but most of it is subjective.

Personally, I kinda agree with @PendaGTP: I prefer better, clear, lean code than "whatever-code" plus documentation. In fact, if the code is obscure, or convoluted, sometimes the docstring can even make things worse. IMHO, docstrings should be used for functions/classes/methods that are reused AND there should be a process to automatically render (pydoc) and publish them.

Bringing that to the context of an Ansible collection, then we are talking primarily about module_utils and plugin_utils (possibly more cases).

But of course, my point of view is very subjective and it relies on one of the two most difficult things in Computer Science: naming things.

IamLunchbox:
I think this is settled then. I already changed ruff.toml to exclude all directores in modules, which are not `_utils`.

@Thulium-Drake this pr is ready.

@PendaGTP Since change detection is enabled you can "undraft" your PRs now :)

Thulium-Drake:
Thanks for working on this! And thanks for participating in the discussion :rocket: :grin:

ruff.toml
```diff
@@ -3,11 +3,12 @@
 # SPDX-FileCopyrightText: 2025 Felix Fontein <felix@fontein.de>
 
 line-length = 120
+include = ["tests/**/*.py", "plugins/**/*.py"]
```
felixfontein, line 6: I didn't look at this close enough before, but this basically disables linting in CI with `ruff check` since it doesn't match any files in CI.

Also please note that restricting paths here is totally incompatible with how antsibull-nox choses paths (especially with change detection). antsibull-nox uses its `code_flies` setting in https://docs.ansible.com/projects/antsibull-nox/config-file/#global-settings.

PR #279, opened 2026-02-06
Format all collection files according to the current ruff config

IamLunchbox:
This PR applies the ruff configuration from #277 and formats + autofixes all files accordingly. 

Addtionally, hidden in this PR is a setting to set the ruff  version to version 3.7, to concur with the guaranteed python compatibility in the readme.

### DIFF
--- a/plugins/doc_fragments/attributes.py
+++ b/plugins/doc_fragments/attributes.py
@@ -1,9 +1,6 @@
-
 # Copyright (c) Ansible Project
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 # SPDX-License-Identifier: GPL-3.0-or-later
-
-
 
 
 class ModuleDocFragment:
--- a/plugins/doc_fragments/proxmox.py
+++ b/plugins/doc_fragments/proxmox.py
@@ -3,8 +3,6 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 
-
-
 class ModuleDocFragment:
     # Common parameters for Proxmox VE modules
     DOCUMENTATION = r"""
--- a/plugins/module_utils/_filelock.py
+++ b/plugins/module_utils/_filelock.py
@@ -6,7 +6,6 @@
 # This has been vendored from ansible.module_utils.common.file. This code has been removed from there for ansible-core 2.16.
 
 
-
 import fcntl
 import os
 import stat
--- a/plugins/module_utils/proxmox.py
+++ b/plugins/module_utils/proxmox.py
@@ -4,7 +4,6 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 
-
 import traceback
 from time import sleep
 
--- a/plugins/module_utils/proxmox_sdn.py
+++ b/plugins/module_utils/proxmox_sdn.py
@@ -4,7 +4,6 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 
-
 from typing import Dict, List
 
 from ansible_collections.community.proxmox.plugins.module_utils.proxmox import (
--- a/plugins/module_utils/version.py
+++ b/plugins/module_utils/version.py
@@ -1,11 +1,7 @@
-
 # Copyright (c) 2021, Felix Fontein <felix@fontein.de>
 # GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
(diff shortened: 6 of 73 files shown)


IamLunchbox:
@felixfontein This PR changed all files in this repo and linting should fail, I think? It does at least when I do run `ruff check`.

Do you have an idea why linting passes? 

In #227 the tests fail because formatting runs before linting and linting is aborted if formatting fails, so that hid the issue from me. 
```
nox > Command ruff format --check --config ruff.toml plugins tests/unit noxfile.py failed with exit code 1
nox > Session formatters failed.
nox > Running session codeqa
nox > Reusing existing virtual environment at .nox/codeqa.
nox > cd /home/runner/work/community.proxmox/community.proxmox/.nox/codeqa/collection-root
nox > ruff check --no-respect-gitignore --output-format=json --config /home/runner/work/community.proxmox/community.proxmox/ruff.toml ansible_collections/community/proxmox/plugins ansible_collections/community/proxmox/tests/unit ansible_collections/community/proxmox/noxfile.py
noxfile.py:9:1: [I001] Import block is un-sorted or un-formatted
                       Organize imports
nox > Session codeqa aborted: Ruff check failed.
nox > Running session antsibull-nox-config
nox > /opt/hostedtoolcache/Python/3.14.2/x64/bin/python /opt/hostedtoolcache/Python/3.14.2/x64/lib/python3.14/site-packages/antsibull_nox/data/antsibull-nox-lint-config.py --data .nox/antsibull-nox-config/tmp/antsibull-nox-lint-config-data.json
nox > Session antsibull-nox-config was successful.
nox > Running session lint
nox > Session lint aborted: Prerequisite session formatters was not successful.
```

felixfontein:
@IamLunchbox in your ruff.toml you have `include = ["tests/**/*.py", "plugins/**/*.py"]`, which works when running ruff from your collection's root, but **not** when running ruff with `ansible_collections/community/proxmox/...`, which antsibull-nox does.

You need to fix the paths in your ruff.toml to either use both variants (if you also want to run `ruff check` directly), or just with the paths antsibull-nox uses (if you don't want to run `ruff check` directly).

IamLunchbox:
I see, so I'd need to remove the `include` directive and extend the `[lint.per-file-ignores]` with `"!ansible_collections/community/proxmox/plugins/*_utils/*.py" = ["D"]`?

I looked through the ci-configuration and the antsibull-nox docs, where and when does the linking / the checkout of this repository into the `ansible_collections` tree actually happen? Maybe I am looking at the wrong documentation, but I could not find a reference of that behavior. :thinking:

felixfontein:
> I see, so I'd need to remove the `include` directive and extend the `[lint.per-file-ignores]` with `"!ansible_collections/community/proxmox/plugins/*_utils/*.py" = ["D"]`?

You don't need to remove the `include` directive (you could also extend it), but I think it's cleaner to not have it there.

> I looked through the ci-configuration and the antsibull-nox docs, where and when does the linking / the checkout of this repository into the `ansible_collections` tree actually happen? Maybe I am looking at the wrong documentation, but I could not find a reference of that behavior. 🤔

It's not explicitly documented, but it's the only way to run `ruff check` correctly in all cases, since some `ruff check` rules do not work fully correct unless you have the correct directory structure (like import ordering).

The same is true when running isort, pylint, or mypy. (black, ruff format, and flake8 don't care about imports in this way, at least to my knowledge.)

IamLunchbox:
I solved it in #277 by removing the include directive and using `!**/*_utils/*.py` to only activate pydocstyle for utils files. And it works with `ruff check` **and** `uv run noxfile.py -e lint`.

Thank you felix!

Thulium-Drake:
I resolved the conflict by adding the incoming change (`!**/*_utils/*.py`), please check the code and let me know if it's OK now :)

IamLunchbox:
- `ruff format` passes now
- Somehow the ruff check for imports and dunder ordering was also listed under `ruff format` in ci
- applied dunder and import ordering first, then autofixes, which were deemed safe by ruff

Thulium-Drake:
So I'm guessing all of the errors that are mentioned in the CI job are everything that is in the repo that doens't conform?

Will it only check changed files on other PRs?

If so, let's merge this, so we can start merging the other PRs and start the refactoring on the modules :-)

IamLunchbox:
Yes, since every file is touched lint should fail as well. 

If the change detection works: i do not know for sure, it should.

felixfontein:
> If the change detection works: i do not know for sure, it should.

It should, but please ping me in case it doesn't, that would be a bug that should get fixed.

Thulium-Drake:
Ok, cool, well, let's see :-)

Thulium-Drake:
Thanks for working on this! :rocket:

PR #291, opened 2026-02-22
style(test_proxmox_vm_info): fix ruff rule PLR2004 (duplicate assert)

PendaGTP:
##### SUMMARY

Remove these assertions since it's redondant with `assert result["proxmox_vms"] == expected_output`

Another step toward fixing the remaining Ruff lint errors.

This one addresses the Ruff rule [PLR2004](https://docs.astral.sh/ruff/rules/magic-value-comparison/) (which fixes 2 lint issues) on file `test_proxmox_vm_info.py`.

First commit remove redondant assertions that cause lint issues. The second one remove remaining redondant assertions (not triggered by lint rule).


Thulium-Drake:
LGTM, thanks for catching it! ;-)

PR #292, opened 2026-02-23
refactor(acl): consolidate acl module

PendaGTP:
##### SUMMARY

This PR consolidate the ACL module.

Bellow main changes:

- lint/format/consistent module doc
- set default state to present instead of absent and mark it as optional
- simplify params validation
- consolidate create/delete methods

This works is related to other changes/refactor we have done.

Note: order of PR merged is not important to me, I will update and rebase when changes merged

EDIT: I will add check/diff mode and update tests after, I did'nt introduce it here, to keep the PR smaller

##### COMPONENT NAME

`proxmox_access_acl`

### DIFF
--- a/tests/unit/plugins/modules/test_proxmox_access_acl.py
+++ b/tests/unit/plugins/modules/test_proxmox_access_acl.py
@@ -15,7 +15,6 @@ from ansible_collections.community.internal_test_tools.tests.unit.plugins.module
     set_module_args,
 )
 
-import ansible_collections.community.proxmox.plugins.module_utils.proxmox as proxmox_utils
 from ansible_collections.community.proxmox.plugins.modules import proxmox_access_acl
 
 ACE = {
@@ -36,7 +35,6 @@ API = {
 class TestProxmoxAccessACLModule(ModuleTestCase):
     def setUp(self):
         super().setUp()
-        proxmox_utils.HAS_PROXMOXER = True
 
         self.connect_mock = patch(
             "ansible_collections.community.proxmox.plugins.module_utils.proxmox.ProxmoxAnsible._connect"


plugins/modules/proxmox_access_acl.py
```diff
@@ -99,65 +95,97 @@
...
+            self._put_acl(**payload)
+        return True
 
 
 def run_module():
     module_args = proxmox_auth_argument_spec()
 
     acl_args = dict(
-        state=dict(choices=["present", "absent"], required=True),
+        state=dict(choices=["present", "absent"], default="present"),
```
Thulium-Drake, line 188: As this _technically_ is a change of a default value, I think we should do the same as for #256 and merge this for 2.0.0

@IamLunchbox Do you agree?

PendaGTP: Sorry I missed to hightlight this change more clearly.

I added a changelog fragment with type `breaking_changes`, let me know we should rename/update it to something else.

plugins/modules/proxmox_access_acl.py
```diff
@@ -9,54 +9,50 @@
...
+    description:
+      - Access Control Path.
+    type: str
+  roleid:
+    description:
+        - The name of the role.
+    type: str
+  type:
+    description:
+        - type of access control
```
IamLunchbox:
```suggestion
        - Type of access control.
```

plugins/modules/proxmox_access_acl.py
```diff
@@ -99,65 +95,97 @@
...
 
-class ProxmoxAccessACLAnsible(ProxmoxAnsible):
-    def _get(self):
-        acls = self.proxmox_api.access.acl.get()
-        return acls
+def _ace_matches(ace, desired):
+    if ace["path"] != desired["path"]:
+        return False
+    roleid = desired.get("roleid")
+    if roleid is not None and ace["roleid"] != roleid:
```
IamLunchbox:
```suggestion
    if roleid and ace["roleid"] != roleid:
```

plugins/modules/proxmox_access_acl.py
```diff
@@ -99,65 +95,97 @@
...
+        return False
+    ace_type = desired.get("type")
+    if ace_type is not None and ace["type"] != ace_type:
```
IamLunchbox:
```suggestion
    if ace_type and ace["type"] != ace_type:
```

plugins/modules/proxmox_access_acl.py
```diff
@@ -99,65 +95,97 @@
...
+        return False
+    ugid = desired.get("ugid")
+    if ugid is not None and ace["ugid"] != ugid:
```
IamLunchbox:
```suggestion
    if ugid and ace["ugid"] != ugid:
```

plugins/modules/proxmox_access_acl.py
```diff
@@ -99,65 +95,97 @@
...
+        return False
+    propagate = desired.get("propagate")
+    if propagate is not None:
```
IamLunchbox:
```suggestion
    if propagate:
```

plugins/modules/proxmox_access_acl.py
IamLunchbox: Since we need to create a major release, I think it does not hurt to declare the breaking change.

PendaGTP:
@IamLunchbox @Thulium-Drake Thanks for review!

I have also rebased.

Done :heavy_check_mark:

IamLunchbox:
Thank you for your help :)

changelogs/fragments/292-set-state-present-as-default-on-acl-module.yml
```diff
@@ -0,0 +1,3 @@
+---
+breaking_changes:
+  - proxmox - set ``state`` as not ``optional`` and assign default value ``present`` (https://github.com/ansible-collections/community.proxmox/pull/292).
```
felixfontein, line 3: I think it should have been "not required", and not "not optional" :) That's also not a breaking change, since any previous use of the parameter is still valid.

PendaGTP: @felixfontein Sorry, should I do anything to "revert" that? Or it too late?

IamLunchbox: Don't worry :)

You can just create a new PR to edit this fragment. It will be parsed on release.

PendaGTP: @IamLunchbox 

This fragment seems to be already added to `changelog.md` on the 1.6.0 release :/ Is it still possible to edit/revert it?

felixfontein: It's already part of a release, but you can still edit it here: https://github.com/ansible-collections/community.proxmox/blob/main/changelogs/changelog.yaml#L304-L306 Afterwards run `antsibull-changelog generate` to update CHANGELOG.md and CHANGELOG.rst.

PendaGTP: @felixfontein Thanks for your help, done here https://github.com/ansible-collections/community.proxmox/pull/370

Issue #280, opened 2026-02-11
Proposal: Add proxmox_domain_realm module


PendaGTP:
## SUMMARY

Add a new module to manage PVE authentication realms.

I'm currently implementing this module and would like to get feedback before opening a PR. I also have several design questions (see below).

### Current implementation proposal

A generic module named `proxmox_domain_realm` to manage all realm types (ad, ldap, openid, ...).

Module documentation:

```yml
module: proxmox_domain_realm
short_description: Management of authentication realms for Proxmox VE cluster
description:
  - Create, update or delete Proxmox VE authentication realms.
author:
  - Clément Cruau (@PendaGTP)
version_added: "1.6.0"
attributes:
  check_mode:
    support: full
  diff_mode:
    support: none
options:
  realm:
    description:
      - The realm name.
    type: str
    aliases: ["name"]
    required: true
  state:
    description:
      - Desired state of the realm.
    choices: ["present", "absent"]
    default: present
    type: str
  type:
    description:
      - The realm type.
      - Only required when O(state=present).
    type: str
    choices: ["ad", "ldap", "openid", "pam", "pve"]
  options:
    description:
      - Realm configuration options.
      - See U(https://pve.proxmox.com/wiki/Manual:_datacenter.cfg) and L(Authentication Realms,https://pve.proxmox.com/pve-docs/chapter-pveum.html).
      - The entire value is masked in logs.
    type: dict

seealso:
  - module: community.proxmox.proxmox_domain_info
    description: Retrieve information about Proxmox VE authentication realms.

extends_documentation_fragment:
  - community.proxmox.proxmox.actiongroup_proxmox
  - community.proxmox.proxmox.documentation
  - community.proxmox.attributes
```

Examples:

```yml
- name: Update the comment on the pam realm
  community.proxmox.proxmox_domain_realm:
    api_host: node1
    api_user: root@pam
    api_password: password
    realm: pam
    type: pam
    options:
      comment: Updated PAM authentication comment

- name: Create a new Proxmox VE openid realm
  community.proxmox.proxmox_domain_realm:
    api_host: node1
    api_user: root@pam
    api_password: password
    realm: openid
    type: openid
    options:
      comment: OpenID authentication
      client-id: 1234567890
      client-key: 1234567890
      issuer-url: https://example.com/issuer

- name: Delete a Proxmox VE openid realm
  community.proxmox.proxmox_domain_realm:
    api_host: node1
    api_user: root@pam
    api_password: password
[...]

PR #293, opened 2026-02-23
Domain

teslamania31:
##### SUMMARY
As discussed in #280 

Add new modules to manage realm (ldap, ad and openid):
 - proxmox_domain : create, update and delete realm
 - proxmox_domain_sync : sync an ldap/ad domain

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
 - proxmox_domain
 - proxmox_domain_sync
 
##### ADDITIONAL INFORMATION
Let me know if this looks good to you, and don’t hesitate to suggest any changes :)

### DIFF
--- a/plugins/modules/proxmox_domain.py
+++ b/plugins/modules/proxmox_domain.py
@@ -22,7 +22,7 @@ attributes:
 options:
     openid_acr_values:
         description:
-            - Specifies the Authentication Context Class Reference values that the Authorization Server is being requested to use for the Authentication Request.
+            - Defines the Authentication Context Class Reference values requested from the Authorization Server for the Authentication Request.
             - Supported for O(type=openid).
         type: str
     openid_autocreate:


plugins/modules/proxmox_domain.py
```diff
@@ -0,0 +1,526 @@
...
+            choices: ['users', 'groups', 'both']
+            type: str
+          enable_new:
+            description: Enable creation of new users.
+            type: bool
+          remove_vanished:
+            description:
+              - A semicolon-separated list of things to remove when they or the user vanishes during a sync.
+              - The following values are possible
+              - entry removes the user/group when not returned from the sync.
```
Thulium-Drake: For readability's sake, I think it's good to wrap the values in (I think?) `O(stuff)`, that should make them stand out a bit more in the rendered docs

plugins/modules/proxmox_domain.py
```diff
@@ -0,0 +1,526 @@
...
+    api_user: "root@pam"
+    api_password: secret
+    realm: "openid"
+    state: present
+    type: "openid"
+    client_id: idoftheclient
+    client_key: keyoftheclient
+    issuer_url: "https://example.test/openid-server"
+
+- name: Del domain
```
Thulium-Drake:
```suggestion
- name: Remove domain
```

I'm nitpicking now, but this is more complete :-)

plugins/modules/proxmox_domain.py
```diff
@@ -0,0 +1,526 @@
...
+
+extends_documentation_fragment:
+  - community.proxmox.proxmox.actiongroup_proxmox
+  - community.proxmox.proxmox.documentation
+  - community.proxmox.attributes
+  - community.proxmox.attributes.info_module
+"""
+
+
+EXAMPLES = r"""
```
Thulium-Drake, line 232: Nice examples, very comprehensive!

plugins/modules/proxmox_domain.py
```diff
@@ -0,0 +1,526 @@
...
+            - Supported for O(type=openid).
+        type: str
+    autocreate:
+        description:
+            - Automatically create users if they do not exist.
+            - Supported for O(type=openid).
+        type: bool
+    base_dn:
+        description:
+            - Ldap base domain name.
```
Thulium-Drake:
```suggestion
            - LDAP base domain name.
```

This is mostly for consistency, can you do a find-replace for `Ldap` -> `LDAP` in the docs part? :-) Same for AD

Oh, this is not for values, I know those are lowercase 😊 

Thanks!

plugins/modules/proxmox_domain_sync.py
```diff
@@ -0,0 +1,139 @@
...
+# Copyright (c) 2026, (@teslamania) <nicolas.vial@protonmail.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+
+DOCUMENTATION = r"""
+module: proxmox_domain_sync
+version_added: 1.6.0
+short_description: Sync realms.
+description: Sync domain realms, ldap or ad.
```
Thulium-Drake:
```suggestion
description: Sync domain realms, LDAP or AD.
```

plugins/modules/proxmox_domain_sync.py
```diff
@@ -0,0 +1,139 @@
...
+author: Vial Nicolas (@teslamania)
+attributes:
+  check_mode:
+    support: full
+  diff_mode:
+    support: none
+
+options:
+    enable_new:
```
Thulium-Drake, line 21: I don't have access to an environment that uses AD or LDAP, but is creating users a part of this API action? If not, this option should only be on the other module

plugins/modules/proxmox_domain_sync.py
```diff
@@ -0,0 +1,139 @@
...
+        type: str
+    realm:
+        description: Authentication domain ID.
+        required: true
+        type: str
+    remove_vanished:
+        description:
+            - A semicolon-separated list of things to remove when they or the user vanishes during a sync.
+            - The following values are possible
+            - entry removes the user/group when not returned from the sync.
```
Thulium-Drake: Same as my comment on the other module

plugins/modules/proxmox_domain_sync.py
```diff
@@ -0,0 +1,139 @@
...
+
+extends_documentation_fragment:
+  - community.proxmox.proxmox.actiongroup_proxmox
+  - community.proxmox.proxmox.documentation
+  - community.proxmox.attributes
+  - community.proxmox.attributes.info_module
+"""
+
+EXAMPLES = r"""
+- name: Sync ldap domain
```
Thulium-Drake:
```suggestion
- name: Sync LDAP domain
```

tests/unit/plugins/modules/test_proxmox_domain.py
```diff
@@ -0,0 +1,340 @@
+#
+# Copyright (c) 2025, teslamania <nicolas.vial@protonmail.com>
```
Thulium-Drake, line 2:
```suggestion
# Copyright (c) 2026, teslamania <nicolas.vial@protonmail.com>
```

;-)

tests/unit/plugins/modules/test_proxmox_domain_sync.py
```diff
@@ -0,0 +1,116 @@
+#
+# Copyright (c) 2025, teslamania <nicolas.vial@protonmail.com>
```
Thulium-Drake, line 2:
```suggestion
# Copyright (c) 2026, teslamania <nicolas.vial@protonmail.com>
```

plugins/modules/proxmox_domain_sync.py
teslamania31: Yes, you can run a sync and override the default values of these three parameters (just like in the UI).
For example, this option enables new users matching the AD or LDAP filter to be created in Proxmox.

plugins/modules/proxmox_domain.py
```diff
@@ -0,0 +1,527 @@
...
+#
+# Copyright (c) 2026, (@teslamania) <nicolas.vial@protonmail.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+
+DOCUMENTATION = r"""
+module: proxmox_domain
+version_added: 1.6.0
+short_description: Manage realms.
```
IamLunchbox:
```suggestion
short_description: Manage authentication realms.
```

plugins/modules/proxmox_domain.py
```diff
@@ -0,0 +1,527 @@
...
+description:
+  - Add, modify or delete domain realms.
```
IamLunchbox:
```suggestion
  - Add, modify or delete domain authentication realms.
```

Maybe link to the pve user management docs as well?

plugins/modules/proxmox_domain.py
```diff
@@ -0,0 +1,527 @@
...
+        description:
+            - LDAP filter for group sync.
+            - Supported for O(type=ldap) and O(type=ad).
+        type: str
+    group_name_attr:
+        description:
+            - LDAP group attribute name.
+            - Supported for O(type=ldap) and O(type=ad).
+        type: str
+    mode:
```
IamLunchbox: I wonder if it would be easier if these as similar options should be preceded by the setting it applies to. That could help distinguish, where it is applicable. And for ldap/ad we could use an alias.

plugins/modules/proxmox_domain.py
```diff
@@ -0,0 +1,527 @@
...
+        description:
+            - Specifies the scopes (user details) that should be authorized and returned, for example 'email' or 'profile'.
+            - Supported for O(type=openid).
+        type: str
+    server1:
+        description:
+            - Server ip address or dns name.
+            - Required when O(type=ldap) or O(type=ad).
+        type: str
+    server2:
```
IamLunchbox:
```suggestion
    primary_server:
        description:
            - Server ip address or dns name.
            - Required when O(type=ldap) or O(type=ad).
        type: str
    secondary_server:
```

plugins/modules/proxmox_domain.py
```diff
@@ -0,0 +1,527 @@
...
+            type: bool
+          remove_vanished:
+            description:
+              - A semicolon-separated list of things to remove when they or the user vanishes during a sync.
+              - The following values are possible
+              - C(remove_vanished=acl) removes acls when the user/group is not returned from the sync.
+              - C(remove_vanished=properties) removes the set properties on existing user/group that do not appear in the source (even custom ones).
+              - C(remove_vanished=entry) removes the user/group when not returned from the sync.
+              - Instead of a list it also can be C(remove_vanishe=none).
+              - Exemple C(remove_vanished="acl;properties;entry")
```
IamLunchbox:
```suggestion
              - Semicolon-separated list of things to remove when they or the user vanishes during a sync.
              - The following values are possible:
              - C(remove_vanished=acl) removes acls when the user/group is not returned from the sync.
              - C(remove_vanished=properties) removes the set properties on existing user/group that do not appear in the source (even custom ones).
              - C(remove_vanished=entry) removes the user/group when not returned from the sync.
              - Instead of a list it also can be C(remove_vanishe=none).
              - Example C(remove_vanished="acl;properties;entry").
```

plugins/modules/proxmox_domain.py
```diff
@@ -0,0 +1,527 @@
...
+        description:
+            - LDAP user attribute name.
+            - Required when O(type=ldap).
+        type: str
+    user_classes:
+        description:
+            - The object class for user (inetorgperson, posixaccount, person, user).
+            - Supported for O(type=ldap) and O(type=ad).
+        type: str
+    verify:
```
IamLunchbox: Maybe this option should indicate in its name, that the back end connection is verified? Otherwise could be confused with the general `validate_certs` from the proxmox auth?

And the docs specify, that the option "capath" must be set as well. It is in the api spec, but not here? It defaults to `/etc/ssl/certs`, maybe we should at least document, that we keep the default? I don't know if there is a scenario, where users set up LDAP and their cert does not reside in there.

plugins/modules/proxmox_domain.py
```diff
@@ -0,0 +1,527 @@
...
+            - LDAP group attribute name.
+            - Supported for O(type=ldap) and O(type=ad).
+        type: str
+    mode:
+        description:
+            - LDAP protocol mode.
+            - Supported for O(type=ldap) and O(type=ad).
+        choices: ['ldap', 'ldaps', 'ldap+starttls']
+        type: str
+    password:
```
IamLunchbox: This is pretty much the same as client_key, isn't it? Maybe this could be a good alias as well?

Or maybe `client_secret` and password, client key? I think in Oauth the terms `Client ID` and `Client Secret` are already the convention.

plugins/modules/proxmox_domain.py
```diff
@@ -0,0 +1,527 @@
...
+        description:
+            - Specifies whether the Authorization Server prompts the End-User for reauthentication and consent.
+            - Supported for with O(type=openid).
+        type: str
+    query_userinfo:
+        description:
+            - Enables querying the userinfo endpoint for claims values.
+            - Supported for with O(type=openid).
+        type: bool
+    realm:
```
IamLunchbox, line 144: Maybe clarify, that this is an arbitrary string which is used to identify the login realm in proxmox. Or rename? `login_realm`? `pve_realm`?

In the end, this is the identifier of the entity, which is edited.

plugins/modules/proxmox_domain.py
```diff
@@ -0,0 +1,527 @@
...
+        domains = self.proxmox_api.access.domains.get()
+        return realm in [item["realm"] for item in domains]
+
+    def get_domain(self, realm):
+        return self.proxmox_api.access.domains.get(realm)
+
+    def is_equal(self, params, current):
+        return all(
+            not (k not in ["password", "realm"] and (k not in current or v != current[k])) for k, v in params.items()
+        )
```
IamLunchbox:
```suggestion
    def is_equal(self, params, current):
        return all(
             (k in current and v != current[k]) or k in ("password", "realm") for k, v in params.items()
        )
```

IamLunchbox:
```suggestion
    def is_equal(self, params, current):
        return all(
            params[k] == current.get(k) for k in params if k not in ("password", "realm")
        )
```

This should be functionally the same and easier to digest :)

IamLunchbox (commented):
Some comments, technically the PR looks fine to me 👍 
Maybe @Thulium-Drake can chime in on my option naming remarks? Maybe I am too strict in my considerations? Some other builtin modules also have options, which are only applied on one type 🤔

plugins/modules/proxmox_domain.py
Thulium-Drake: That would work and make it easy for the endusers to use :-)

So yeah, good suggestion!

plugins/modules/proxmox_domain.py
Thulium-Drake: Well, not really in an LDAP context, _functionally_ yes, but LDAP doesn't do `client_secrets`, all fields I've seen so far in LDAP servers is `password` :-)

plugins/modules/proxmox_domain.py
Thulium-Drake: If we're renaming this, let's stick to what Proxmox uses in their docs -> `authentication_realm`

https://pve.proxmox.com/pve-docs/chapter-pveum.html#pveum_authentication_realms

plugins/modules/proxmox_domain.py
Thulium-Drake: `validate_ad_certs` with an alias to `validate_ldap_certs`?

plugins/modules/proxmox_domain.py
teslamania31: Just to confirm: should every parameter that applies to a specific type be prefixed with that type?

For example:
Parameters only for AD: case_sensitive, domain → ad_case_sensitive, ad_domain
Parameters only for LDAP: base_dn, user_attr → ldap_base_dn, ldap_user_attr
Parameters shared by AD and LDAP (bind_dn, filter, mode, port, password, etc.) ldap_.. with an alias ad_..
And similarly for OpenID: issuer_url → openid_issuer_url, etc.

plugins/modules/proxmox_domain.py
teslamania31: the colon after "The following values are possible" breaks the sanity check

plugins/modules/proxmox_domain.py
Thulium-Drake: Yes, that's what I'd do. This will make it very clear when reading the playbooks what it will do.

It's more aliases, sure, but I think it makes it easier to manage :-)

plugins/modules/proxmox_domain.py
teslamania31: now ldap_primary_server with an alias ad_primary_server and same for secondary

plugins/modules/proxmox_domain.py
teslamania31: In fact it is the pasword of the ldap_bind_dn. It is used to sync the users/groups

plugins/modules/proxmox_domain.py
teslamania31: I prefer not rename it if it is possible. I clarified the description.

plugins/modules/proxmox_domain.py
```diff
@@ -0,0 +1,551 @@
...
+attributes:
+  check_mode:
+    support: full
+  diff_mode:
+    support: none
+
+options:
+    openid_acr_values:
+        description:
+            - Specifies the Authentication Context Class Reference values that theAuthorization Server is being requested to use for the Auth Request.
```
IamLunchbox:
```suggestion
            - Specifies the Authentication Context Class Reference values that the Authorization Server is being requested to use for the Authentication Request.
```

Thulium-Drake:
Thanks for your work (and your patience ;-) )!

PR #295, opened 2026-02-24
Update readme and contributors guide

IamLunchbox:
This PR updates the README by reordering and streamlining the content. First and foremost aim was provide all the necessary information right away.

And the contribution guide was overhauled as well, to provide more information how developers can contribute and test their code. The guide was heavily copied from community.general but then edited to fit our needs .

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,30 +6,242 @@ SPDX-License-Identifier: GPL-3.0-or-later
 
 # Contributing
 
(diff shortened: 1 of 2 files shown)


Thulium-Drake:
Thanks a lot for writing this, it basically addresses all the hurdles one might need to jump through when wanting to make a contribution! :-)

And when writing the example changelogs I completely missed that there's docs for that :hand_over_mouth:

PR #307, opened 2026-03-02
perf(pool): check if pool exists by using the poolid

PendaGTP:
##### SUMMARY

Improve performance of the module `proxmox_pool` by checking if the pool exists using the ID instead of retrieving all pools.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

`proxmox_pool`


Thulium-Drake:
Thanks for this work! :-)

PR #309, opened 2026-03-02
fix(proxmox_user_info): return empty user list when user not found

emmanueljamet:
##### SUMMARY
fix proxmox_user_info returning an error in case the queried user doesn't exist.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
proxmox_user_info

##### ADDITIONAL INFORMATION
Docs already states that `proxmox_user_info` always returns `proxmox_users` but can be empty, with this PR it's now true :)

Also, not finding a user is just an info, quite unexpected to be forced to handle an error in that case, IMHO.

from
```fatal: [localhost]: FAILED! => {"changed": false, "msg": "User 'some-user@pve' does not exist"}```
to
```
ok: [localhost] => {
    "proxmox_user": {
        "changed": false,
        [...]
        "failed": false,
        "proxmox_users": []
    }
}
```

### DIFF
--- a/plugins/modules/proxmox_user_info.py
+++ b/plugins/modules/proxmox_user_info.py
@@ -153,6 +153,8 @@ proxmox_users:
 """
 
 
+import traceback
+
 from ansible.module_utils.basic import AnsibleModule
 
 from ansible_collections.community.proxmox.plugins.module_utils.proxmox import (
@@ -160,8 +162,6 @@ from ansible_collections.community.proxmox.plugins.module_utils.proxmox import (
     proxmox_auth_argument_spec,
     proxmox_to_ansible_bool,
 )
-
-import traceback
 
 try:
     import proxmoxer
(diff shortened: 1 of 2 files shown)


IamLunchbox:
Thank you for your bugfix! Could you try to narrow down the first except statement to the exception thrown in your example, and then another `except Exception as e:` to get another failure message, if something else goes wrong? :)

Good catch, though! Your solution is well fitting (:

emmanueljamet:
> Thank you for your bugfix! Could you try to narrow down the first except statement to the exception thrown in your example, and then another `except Exception as e:` to get another failure message, if something else goes wrong? :)
> 
> Good catch, though! Your solution is well fitting (:

Thanks for the quick review!

Sure, done.

emmanueljamet:
I've also added unit tests for this module, hope that helps

Thulium-Drake:
Looks good, thanks for this!

Issue #237, opened 2025-12-15
Add Ansible info module for Proxmox cluster status and quorum


michaelwdombek:
##### SUMMARY

Add a read-only info module to retrieve Proxmox cluster status information via the API, including cluster nodes, quorum state, and cluster configuration details.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

community.proxmox.proxmox_cluster_status_info

##### ADDITIONAL INFORMATION

Currently there is no supported way in the community Proxmox modules to retrieve the overall cluster status, such as the list of nodes in a cluster or whether the cluster is quorate. This information is available via the Proxmox API endpoint [GET /cluster/status](https://pve.proxmox.com/pve-docs/api-viewer/index.html#/cluster/status), but cannot be consumed easily in Ansible workflows.

This creates gaps in automation, especially for:
- validating cluster health and quorum state,
- cluster creation and bootstrap workflows where cluster_join_info is insufficient, 
- conditional logic depending on node availability.

```yaml
- name: Get cluster status
  community.proxmox.proxmox_cluster_status_info:
    api_host: proxmox1
    api_user: root@pam
    api_password: "{{ password | default(omit) }}"
    api_token_id: "{{ token_id | default(omit) }}"
    api_token_secret: "{{ token_secret | default(omit) }}"
  register: cluster_status

- name: Display cluster quorum status
  ansible.builtin.debug:
    msg: "Cluster has quorum: {{
      cluster_status.cluster_status
      | selectattr('type', 'equalto', 'cluster')
      | map(attribute='quorate')
      | first
    }}"

```

PR #313, opened 2026-03-08
Release 1.6.0

Thulium-Drake:
##### SUMMARY

Release 1.6.0

##### ISSUE TYPE

- Release Pull Request

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,44 +2,52 @@
 
 **Topics**
 
-- <a href="#v1-5-0">v1\.5\.0</a>
+- <a href="#v1-6-0">v1\.6\.0</a>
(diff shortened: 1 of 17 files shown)


Thulium-Drake:
Rebased to pull in #314 and #237

nickbroon:
As this contained "Breaking Changes" should the version have been 2.0.0 instead of 1.6.0? The semantic versioning rules would seem to suggest that: 

https://docs.ansible.com/projects/antsibull-changelog/changelogs/

> breaking_changes
>
> This (new) category should list all changes to features which absolutely require attention from users when upgrading, because an existing behavior is changed. This is mostly what Ansible's Porting Guide used to describe. **This section should only appear in a initial major release (x.0.0) according to semantic versioning**

PR #316, opened 2026-03-09
fix(pool): support nested poolid

PendaGTP:
##### SUMMARY

Proxmox VE 8.1+ supports nested pools, this PR fixes current implementation which not support it.

Example:

```bash
[ERROR]: Task failed: Module failed: Unable to retrieve pool test/test: 501 Not Implemented: Method 'GET /pools/test/test' not implemented
Origin: poc-pool.yml

8   tasks:
9     - name: Create new Proxmox VE pool
        ^ column 7

fatal: [localhost]: FAILED! => {"changed": false, "msg": "Unable to retrieve pool test/test: 501 Not Implemented: Method 'GET /pools/test/test' not implemented"}
```

Use query parameter form so nested pool IDs with "/" work: `GET /api2/json/pools?poolid=<poolid>`

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

`proxmox_pool`

PR #289, opened 2026-02-22
 Refactor module creation

PendaGTP:
### Summary

Create a new helper `create_proxmox_module()` which adds generic auth args and constraints, and merges in the module-specific args and options

Also some other little standardization related:

- args and options are defined on top of module class
- first two lines of main() are create module and create the specific module instance
- use dict for args specs instead of literal (e.g plugins/modules/proxmox_backup.py)
- fix check mode that were not matching (doc vs module options, e.g plugins/modules/proxmox.py)
- remove duplicate proxmoxer checks (e.g plugins/modules/proxmox_backup_schedule.py)

Tests done using:

```toml
# antsibull-nox.toml
# only changed files
[...]

PR #325, opened 2026-03-14
feat(proxmox_user): add support for tokens

emmanueljamet:
##### SUMMARY
Add support for tokens in `proxmox_user` module

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
`proxmox_user`

##### ADDITIONAL INFORMATION

```
- name: Create Terraform user
  community.proxmox.proxmox_user:
    name: terraform
    password: "{{ password }}"
    [...]
    tokens:
      - tokenid: automation
        comment: "Token used by terraform automation"
        expire: 0
        privsep: False
  register: token_info       # <<< this allows to retrieve token - remember it is only displayed once, at creation
  no_log: true                  # <<< this is recommened if you don't want to leak your token in logs
```

which gives (without the `no_log` option):
```
TASK [Create Terraform user] ***************************************************************************************************************************************************************
changed: [myserver] => {"changed": true, "msg": "Created user terraform@pve", "secrets": {"automation": "6ba9e9ba-53bb-469b-8cbf-7c4a39450356"}, "userid": "terraform@pve"}
```

### DIFF
--- a/plugins/modules/proxmox_user.py
+++ b/plugins/modules/proxmox_user.py
@@ -78,6 +78,29 @@ options:
       - Only for PVE Authentication Realm users.
       - Parameter is ignored when user already exists or O(state=absent).
     type: str
+  tokens:
+    description: List of API tokens associated to the user.
+    type: list
+    elements: dict
+    suboptions:
+      comment:
+        description: Short description of the token.
+        type: str
+      expire:
+        description: Expiration date in seconds since EPOCH. Zero means no expiration.
+        type: int
+        default: 0
+      privsep:
+        description: Describe if the API token is further restricted with ACLs or is fully privileged.
+        type: bool
+        default: true
+      tokenid:
+        description:
+          - Token name.
+          - Case sensitive and must be unique per user.
+        type: str
+        aliases: ["name"]
+        required: true
 
 extends_documentation_fragment:
   - community.proxmox.proxmox.actiongroup_proxmox
@@ -100,6 +123,11 @@ EXAMPLES = r"""
     groups:
       - admins
     password: GoBananas!
+    tokens:
+      - tokenid: MyToken
+        comment: Expires on 2026-01-01 00:00:00
+        expire: 1767222000
+        privsep: false
     lastname: Some Guy
 
 - name: Delete a Proxmox VE user
@@ -117,6 +145,10 @@ userid:
   returned: success
   type: str
   sample: test
+secrets:
+  description: Dictionary of API tokens associated with their secret.
+  returned: success
+  type: dict
 msg:
   description: A short message on what the module did.
   returned: always
@@ -126,6 +158,7 @@ msg:
 
 from ansible_collections.community.proxmox.plugins.module_utils.proxmox import (
     ProxmoxAnsible,
+    ansible_to_proxmox_bool,
     create_proxmox_module,
 )
 
@@ -142,6 +175,19 @@ def module_args():
         lastname=dict(type="str"),
         keys=dict(type="str", no_log=True),
         password=dict(type="str", no_log=True),
+        tokens=dict(
+            type="list",
+            no_log=False,
+            elements="dict",
+            options=dict(
+                tokenid=dict(
+                    type="str", aliases=["name"], no_log=False, required=True
+                ),  # no_log is necessary to pass the CI
+                comment=dict(type="str"),
+                expire=dict(type="int", default=0),
+                privsep=dict(type="bool", default=True),
+            ),
+        ),
         state=dict(default="present", choices=["present", "absent"]),
     )
 
@@ -191,6 +237,82 @@ class ProxmoxUserAnsible(ProxmoxAnsible):
 
         return False
 
+    def _tokens_need_update(self, existing_tokens, new_tokens):
+        """Check if tokens need updating by comparing current vs desired state"""
+        existing_tokens = existing_tokens or {}
+        new_tokens = new_tokens or []
+
+        # If the number of tokens is different, we need an update
+        if len(existing_tokens) != len(new_tokens):
+            return True
+
+        # Check tokens - If any token is different or missing, we consider an update is needed.
+        for token in new_tokens:
+            existing_token = existing_tokens.get(token["tokenid"])
+            if not existing_token:
+                return True
+            if (
+                existing_token.get("comment", "") != (token["comment"] or "")
+                or existing_token.get("expire", 0) != token["expire"]
+                or bool(existing_token.get("privsep")) != token["privsep"]
+            ):
+                return True
+
+        return False
+
+    def create_update_delete_tokens(self, userid: str, tokens: list = None) -> set:
+        result_tokens = {}
+        tokens = tokens or []
+        candidate_token_ids = [t["tokenid"] for t in tokens]
+
+        try:
+            existing_tokens = self.is_user_existing(userid).get("tokens", {}) or {}
+        except AttributeError:
+            self.module.fail_json(
+                changed=False,
+                userid=userid,
+                msg=f"Failed to manage tokens for unknown user with ID {userid}",
+            )
+
+        for token in tokens:
+            try:
+                if token["tokenid"] not in existing_tokens:
+                    resp = (
+                        self.proxmox_api.access.users(userid)
+                        .token(token["tokenid"])
+                        .post(
+                            tokenid=token["tokenid"],
+                            comment=token["comment"],
+                            expire=token["expire"],
+                            privsep=ansible_to_proxmox_bool(token["privsep"]),
+                        )
+                    )
+                    result_tokens[token["tokenid"]] = resp["value"]
+                else:
+                    self.proxmox_api.access.users(userid).token(token["tokenid"]).put(
+                        comment=token["comment"],
+                        expire=token["expire"],
+                        privsep=ansible_to_proxmox_bool(token["privsep"]),
+                    )
+            except Exception as e:
+                self.module.fail_json(
+                    changed=False,
+                    userid=userid,
+                    msg=f"Failed to create or update token '{token['tokenid']}' for user with ID {userid}: {e}",
+                )
+
+        for existing_token in existing_tokens:
+            try:
+                if existing_token not in candidate_token_ids:
+                    self.proxmox_api.access.users(userid).token(existing_token).delete()
+            except Exception as e:
+                self.module.fail_json(
+                    changed=False,
+                    userid=userid,
+                    msg=f"Failed to delete token '{existing_token}' for user with ID {userid}: {e}",
+                )
+        return result_tokens
+
     def create_update_user(  # noqa: PLR0913
         self,
         userid,
@@ -201,6 +323,7 @@ class ProxmoxUserAnsible(ProxmoxAnsible):
         firstname=None,
         groups=None,
         password=None,
+        tokens=None,
         keys=None,
         lastname=None,
     ):
@@ -214,6 +337,7 @@ class ProxmoxUserAnsible(ProxmoxAnsible):
         :param firstname: str, optional - First name of the user
         :param groups: list, optional - Groups that the user should be a member of
         :param password: str, optional - Password of the user, PVE realm only
+        :param tokens: list, optional - API tokens associated to the user
         :param keys: str, optional - 2FA keys for the user
         :param lastname: str, optional - Lastname of the user
         :return: None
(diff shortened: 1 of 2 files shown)


plugins/modules/proxmox_user.py
```diff
@@ -78,6 +78,29 @@
...
+    suboptions:
+      comment:
+        description: Short description of the token.
+        type: str
+      expire:
+        description: Expiration date in seconds since EPOCH. Zero means no expiration.
+        type: int
+        default: 0
+      privsep:
+        description: Describe if the API token is further restricted with ACLs (true) or is fully privileged (false).
```
IamLunchbox:
```suggestion
        description: Describe if the API token is further restricted with ACLs V(true) or is fully privileged V(false).
```

plugins/modules/proxmox_user.py
```diff
@@ -78,6 +78,29 @@
...
+    elements: dict
+    suboptions:
+      comment:
+        description: Short description of the token.
+        type: str
+      expire:
+        description: Expiration date in seconds since EPOCH. Zero means no expiration.
+        type: int
+        default: 0
+      privsep:
```
IamLunchbox, line 93:
```suggestion
      privilege_separation:
```

Just a suggestion, I think this is more readable when you use it in a playbook. But that would require you to convert privilege_separation to privsep at a later point.

plugins/modules/proxmox_user.py
```diff
@@ -78,6 +78,29 @@
...
+        description: Describe if the API token is further restricted with ACLs (true) or is fully privileged (false).
+        type: bool
+        default: true
+      tokenid:
```
IamLunchbox, line 97:
```suggestion
      token_id:
```

plugins/modules/proxmox_user.py
```diff
@@ -148,7 +180,7 @@ def is_user_existing(self, userid):
             else:
                 self.module.fail_json(msg=f"Unable to retrieve user {userid}: {e}")
 
-    def _user_needs_update(self, existing_user, comment, email, enable, expire, firstname, lastname, groups, keys):
+    def _user_needs_update(self, existing_user, comment, email, enable, expire, firstname, lastname, groups, keys):  # noqa: PLR0913
```
IamLunchbox, line 215:
```suggestion
    def _user_needs_update(self, existing_user, comment, email, enable, expire, firstname, lastname, groups, keys): 
```

Please don't exclude these errors unless absolutely necessary. This function problably needs a rewrite or could source this information from self.

plugins/modules/proxmox_user.py
```diff
@@ -173,7 +205,30 @@ def _user_needs_update(self, existing_user, comment, email, enable, expire, firs
...
         return False
 
-    def create_update_user(
+    def _tokens_need_update(self, existing_tokens, new_tokens):
+        """Check if tokens need updating by comparing current vs desired state"""
+        existing_tokens = existing_tokens or {}
+        new_tokens = new_tokens or []
+
+        # If the number of tokens is different, we need an update
+        if len(existing_tokens.keys()) != len(new_tokens):
```
IamLunchbox:
```suggestion
        if len(existing_tokens) != len(new_tokens):
```

plugins/modules/proxmox_user.py
```diff
@@ -173,7 +205,30 @@ def _user_needs_update(self, existing_user, comment, email, enable, expire, firs
...
+        for token in new_tokens:
+            existing_token = existing_tokens.get(token["tokenid"])
+            if not existing_token:
+                return True
+            if (
+                existing_token.get("comment", "") != (token["comment"] or "")
+                or existing_token.get("expire", 0) != token["expire"]
+                or bool(existing_token.get("privsep")) != token["privsep"]
+            ):
+                return True
```
IamLunchbox, line 259:
```suggestion
                existing_token.get("comment", "") != token.get("comment", "")
                or existing_token.get("expire", 0) != token.get("expire")
                or proxmox_to_ansible_bool(existing_token.get("privsep")) != token.get("privsep")
            ):
                return True
```

In review mode I cant add imports, but I think this is probably more what you want, privsep is probably an int between 0 and 1. You need to import it from proxmox.py.

plugins/modules/proxmox_user.py
```diff
@@ -218,6 +277,7 @@ def create_update_user(
                 )
 
             if needs_update:
+                existing_tokens = existing_user.get("tokens") or {}
```
IamLunchbox:
```suggestion
                existing_tokens = existing_user.get("tokens", {})
```

plugins/modules/proxmox_user.py
```diff
@@ -233,7 +293,35 @@ def create_update_user(
...
+                                    expire=token["expire"],
+                                    privsep=(1 if token["privsep"] else 0),
+                                )
+                            )
+                            result_tokens[token["tokenid"]] = resp["value"]
+                        else:
+                            self.proxmox_api.access.users(userid).token(token["tokenid"]).put(
+                                comment=token["comment"],
+                                expire=token["expire"],
+                                privsep=(1 if token["privsep"] else 0),
```
IamLunchbox:
```suggestion
                                privsep=ansible_to_proxmox_bool(token["privilege_separation"]),
```

plugins/modules/proxmox_user.py
```diff
@@ -233,7 +293,35 @@ def create_update_user(
...
+                    for token in tokens:
+                        if len(existing_tokens) == 0 or token["tokenid"] not in existing_tokens:
+                            resp = (
+                                self.proxmox_api.access.users(userid)
+                                .token(token["tokenid"])
+                                .post(
+                                    tokenid=token["tokenid"],
+                                    comment=token["comment"],
+                                    expire=token["expire"],
+                                    privsep=(1 if token["privsep"] else 0),
```
IamLunchbox:
```suggestion
                                    privsep=ansible_to_proxmox_bool(token["privilege_separation"]),
```

plugins/modules/proxmox_user.py
```diff
@@ -233,7 +293,35 @@ def create_update_user(
                         if value is not None:
                             update_params[field] = value
                     self.proxmox_api.access.users(userid).put(**update_params)
-                    self.module.exit_json(changed=True, userid=userid, msg=f"User {userid} updated")
+
+                    result_tokens = {}
+                    for token in tokens:
```
IamLunchbox: Wouldn't it make sense to go into a separate function here?

plugins/modules/proxmox_user.py
```diff
@@ -173,7 +205,30 @@ def _user_needs_update(self, existing_user, comment, email, enable, expire, firs
...
+            if (
+                existing_token.get("comment", "") != (token["comment"] or "")
+                or existing_token.get("expire", 0) != token["expire"]
+                or bool(existing_token.get("privsep")) != token["privsep"]
+            ):
+                return True
+
+        return False
+
+    def create_update_user(  # noqa: PLR0912, PLR0913
```
IamLunchbox:
```suggestion
    def create_update_user(
```

plugins/modules/proxmox_user.py
```diff
@@ -269,7 +357,22 @@ def create_update_user(
...
+            result_tokens = {}
+            for token in tokens:
+                resp = (
+                    self.proxmox_api.access.users(userid)
+                    .token(token["tokenid"])
+                    .post(
+                        tokenid=token["tokenid"],
+                        comment=token["comment"],
+                        expire=token["expire"],
+                        privsep=(1 if token["privsep"] else 0),
```
IamLunchbox:
```suggestion
                        privsep=ansible_to_proxmox_bool(token["privilege_separation"]),
```

plugins/modules/proxmox_user.py
```diff
@@ -305,6 +408,17 @@ def main():
         lastname=dict(type="str"),
         keys=dict(type="str", no_log=True),
         password=dict(type="str", no_log=True),
+        tokens=dict(
+            type="list",
+            no_log=False,
+            elements="dict",
+            options=dict(
+                tokenid=dict(type="str", aliases=["name"], no_log=False, required=True),
```
IamLunchbox:
```suggestion
                token_id=dict(type="str", aliases=["name"], required=True),
```

Or was there a specific point to set `no_log` here?

plugins/modules/proxmox_user.py
```diff
@@ -305,6 +408,17 @@ def main():
...
+                comment=dict(type="str"),
+                expire=dict(type="int", default=0),
+                privsep=dict(type="bool", default=True),
```
IamLunchbox:
```suggestion
                privilege_separation=dict(type="bool", default=True),
```

plugins/modules/proxmox_user.py
emmanueljamet: I used the variable name as defined in the API and the CLI. I think it is easier to use the same vocabulary as the official one

plugins/modules/proxmox_user.py
emmanueljamet: I used the variable name as defined in the API and the CLI. I think it is easier to use the same vocabulary as the official one

plugins/modules/proxmox_user.py
emmanueljamet: Yes, but it is not part of my PR, maybe we should create another ticket for that.
No idea how it passed the CI previously, though.

plugins/modules/proxmox_user.py
emmanueljamet: I did not modify this in my PR, I don't know how it passed the CI so far...

plugins/modules/proxmox_user.py
emmanueljamet: No, no, good catch

plugins/modules/proxmox_user.py
emmanueljamet: Plus, this is the exact naming for `proxmox_user_info` module

emmanueljamet:
> Thank you for your PR! 🚀 I would suggest renaming two options, privsep and tokenid. I highlighted some occurrences, but not all. Especially in your unit tests. When you rename it you might need to fix that as well.
> 
> Additionally, please remove the lint ignores from the functions, these need to be refactored. From a functional point of view: LGTM!

Thanks for the review,

- I created a function and added usage of `proxmox_to_ansible_bool`
- `no_log` parameter is useful to pass the CI (`ERROR: plugins/modules/proxmox_user.py:0:0: no-log-needed: Argument 'tokenid' in argument_spec could be a secret, though doesn't have `no_log` set found in tokens`)
- For the naming, I reused what's defined in proxmox_user_info, that also fits what's defined in the proxmox API and CLI
- I removed the error exclusions ; I don't know why I needed them at some point 🤷🏻

emmanueljamet:
(please also note that a recent change on main branch is breaking the CI because of deprecations of `validate_certs`)

PendaGTP:
> (please also note that a recent change on main branch is breaking the CI because of deprecations of `validate_certs`)

This will be fixed by https://github.com/ansible-collections/community.proxmox/pull/330

Related to this comment: https://github.com/ansible-collections/community.proxmox/pull/316#issuecomment-4031877435

plugins/modules/proxmox_user.py
IamLunchbox: @Thulium-Drake What is your take on this?

Thulium-Drake:
@emmanueljamet Thanks for this work! I have rebased your code, but there's some sanity things left :-)

emmanueljamet:
What should I do with these CI errors?

```
plugins/modules/proxmox_user.py:184:9: [PLR0913] Too many arguments in function definition (9 > 5)
plugins/modules/proxmox_user.py:285:9: [PLR0913] Too many arguments in function definition (11 > 5)
```

It was KO even before I added a parameter, this needs another refactor

plugins/modules/proxmox_user.py
Thulium-Drake: I think the most elegant solution would be to use aliases, I've seen this in more modules (have a look at `community.general.keycloak_*`) where they support both the 'native' API name as well as something 'normal'.

Thulium-Drake:
Yes, we can ignore those for now, these errors are outside the code you've changed.

Also,  #289 was merged, which means that there's some merge conflicts after the new means of setting up the modules has slightly changed.

Can you please had a look and fix the merge conflicts? :rocket:

emmanueljamet:
@Thulium-Drake done, could you check again please?

plugins/modules/proxmox_user.py
emmanueljamet: Yes that's an elegant solution but I think it deserves its own PR to harmonize everywhere

emmanueljamet:
All green, can we merge?

plugins/modules/proxmox_user.py
Thulium-Drake: Sounds like a plan to me :-)

PR #333, opened 2026-03-16
add new module ceph pool

teslamania31:
##### SUMMARY
Add new module to manage ceph pool

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
proxmox_ceph_pool

### DIFF
--- a/plugins/module_utils/proxmox.py
+++ b/plugins/module_utils/proxmox.py
@@ -432,3 +432,14 @@ class ProxmoxAnsible:
             return self.proxmox_api.nodes(node).storage(storage).content().get(content=content, vmid=vmid)
         except Exception as e:
             self.module.fail_json(msg=f"Unable to list content on {node}, {storage} for {content} and {vmid}: {e}")
+
+    def check_node_on_cluster(self, node):
+        """
+        Checks if the node is part of the cluster.
+
+        Args:
+            node(str): searched node.
+        """
+        nodes = self.proxmox_api.cluster.resources.get(type="node")
+        if node not in [item["node"] for item in nodes]:
+            self.module.fail_json(msg=f"Node {node} does not exist in the cluster.")
--- a/plugins/module_utils/proxmox.py
+++ b/plugins/module_utils/proxmox.py
@@ -432,14 +432,3 @@ class ProxmoxAnsible:
             return self.proxmox_api.nodes(node).storage(storage).content().get(content=content, vmid=vmid)
         except Exception as e:
             self.module.fail_json(msg=f"Unable to list content on {node}, {storage} for {content} and {vmid}: {e}")
-
-    def node_in_cluster(self, node):
-        """
-        Checks if the node is part of the cluster.
-
-        Args:
-            node(str): searched node.
-        """
-        nodes = self.proxmox_api.cluster.resources.get(type="node")
-        if node not in [item["node"] for item in nodes]:
-            self.module.fail_json(msg=f"Node {node} does not exist in the cluster.")
--- a/plugins/module_utils/proxmox.py
+++ b/plugins/module_utils/proxmox.py
@@ -433,7 +433,7 @@ class ProxmoxAnsible:
         except Exception as e:
             self.module.fail_json(msg=f"Unable to list content on {node}, {storage} for {content} and {vmid}: {e}")
 
-    def check_node_on_cluster(self, node):
+    def node_in_cluster(self, node):
         """
         Checks if the node is part of the cluster.
 
--- a/plugins/modules/proxmox_ceph_pool.py
+++ b/plugins/modules/proxmox_ceph_pool.py
@@ -20,7 +20,7 @@ attributes:
 
 options:
     add_storages:
-        description: If true, add the new pool to the cluster storage configuration.
+        description: Add the new pool to the cluster storage configuration.
         required: false
         type: bool
     crush_rule:
@@ -40,7 +40,7 @@ options:
         required: true
         type: str
     pg_autoscale_mode:
-        description: The automatic PG scaling mode of the pool.
+        description: The automatic placement groups scaling mode of the pool.
         required: false
         type: str
         choices: ["on", "off", "warn"]
@@ -61,12 +61,17 @@ options:
         description: Number of replicas per object.
         required: false
         type: int
+    timeout:
+        description: Timeout for operations.
+        default: 5
+        required: false
+        type: int
     target_size:
-        description: The estimated target size of the pool for the PG autoscaler.
+        description: The estimated target size of the pool for the placement groups autoscaler.
         required: false
         type: str
     target_size_ratio:
-        description: The estimated target ratio of the pool for the PG autoscaler.
+        description: The estimated target ratio of the pool for the placement groups autoscaler.
         required: false
         type: int
 
@@ -79,7 +84,7 @@ extends_documentation_fragment:
 
 
 EXAMPLES = r"""
-- name: Add a pool ceph
+- name: Add a ceph pool
   community.proxmox.proxmox_ceph_pool:
     api_host: proxmox
     api_user: root@pam
@@ -88,7 +93,7 @@ EXAMPLES = r"""
     name: ceph-pool
     state: present
 
-- name: Add a pool ceph and the storage
+- name: Add a ceph pool and the storage
   community.proxmox.proxmox_ceph_pool:
     api_host: proxmox
     api_user: root@pam
@@ -98,7 +103,7 @@ EXAMPLES = r"""
     state: present
     add_storages: true
 
-- name: Delete a pool ceph
+- name: Delete a ceph pool
   community.proxmox.proxmox_ceph_pool:
     api_host: proxmox
     api_user: root@pam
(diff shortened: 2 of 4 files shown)


plugins/modules/proxmox_ceph_pool.py
```diff
@@ -0,0 +1,265 @@
...
+author: Vial Nicolas (@teslamania)
+attributes:
+  check_mode:
+    support: full
+  diff_mode:
+    support: none
+
+options:
+    add_storages:
+        description: If true, add the new pool to the cluster storage configuration.
```
IamLunchbox:
```suggestion
        description: Add the new pool to the cluster storage configuration.
```

plugins/modules/proxmox_ceph_pool.py
```diff
@@ -0,0 +1,265 @@
...
+    name:
+        description: The name of the pool. It must be unique.
+        required: true
+        type: str
+    node:
+        description: The cluster node name.
+        required: true
+        type: str
+    pg_autoscale_mode:
+        description: The automatic PG scaling mode of the pool.
```
IamLunchbox: What do you think about writing out placement group here as well? (:

plugins/modules/proxmox_ceph_pool.py
```diff
@@ -0,0 +1,265 @@
...
+extends_documentation_fragment:
+  - community.proxmox.proxmox.actiongroup_proxmox
+  - community.proxmox.proxmox.documentation
+  - community.proxmox.attributes
+  - community.proxmox.attributes.info_module
+"""
+
+
+EXAMPLES = r"""
+- name: Add a pool ceph
```
IamLunchbox:
```suggestion
- name: Add a  ceph pool
```

plugins/modules/proxmox_ceph_pool.py
```diff
@@ -0,0 +1,265 @@
...
+  community.proxmox.proxmox_ceph_pool:
+    api_host: proxmox
+    api_user: root@pam
+    api_password: secret
+    node: proxmox
+    name: ceph-pool
+    state: present
+
+- name: Add a pool ceph and the storage
```
IamLunchbox:
```suggestion
- name: Add a ceph pool and storage
```

plugins/modules/proxmox_ceph_pool.py
```diff
@@ -0,0 +1,265 @@
...
+  community.proxmox.proxmox_ceph_pool:
+    api_host: proxmox
+    api_user: root@pam
+    api_password: secret
+    node: proxmox
+    name: ceph-pool
+    state: present
+    add_storages: true
+
+- name: Delete a pool ceph
```
IamLunchbox:
```suggestion
- name: Delete a ceph pool
```

plugins/modules/proxmox_ceph_pool.py
```diff
@@ -0,0 +1,265 @@
...
+        }
+        return params
+
+    def is_equal(self, params, current):
+        return all(params[k] == current.get(k) for k in params if k not in ("add_storages", "node"))
+
+    def check_node(self, node):
+        nodes = self.proxmox_api.cluster.resources.get(type="node")
+        if node not in [item["node"] for item in nodes]:
+            self.module.fail_json(msg=f"Node {node} does not exist in the cluster.")
```
IamLunchbox: This could be a shared function in proxmox.py I suppose :)

plugins/modules/proxmox_ceph_pool.py
```diff
@@ -0,0 +1,265 @@
...
+        node = self.params["node"]
+        name = self.params["name"]
+        self.check_node(node)
+        if self.check_pool(node, name):
+            if self.module.check_mode:
+                self.module.exit_json(changed=True, msg=f"Ceph pool {name} would be deleted.")
+            else:
+                current_task_id = self.proxmox_api.nodes(node).ceph.pool(name).delete()
+                task_success, fail_reason = self.api_task_complete(
+                    node, current_task_id, self.module.params["api_timeout"]
```
IamLunchbox: I don't know if using `api_timeout` as an option here is against the user expectation? You don't want to enable round trip times of 30 seconds, but you may want to wait 30 seconds for a cluster to be built.

And you did not document it. Maybe a user wants to fail and sets a low timeout and then wonders, why the ceoph pools always timeout? 🤔

plugins/modules/proxmox_ceph_pool.py
teslamania31: good idea ! Once it is merged, I could also use it in the other ceph modules

plugins/modules/proxmox_ceph_pool.py
teslamania31: You're right, It was a misuse of the api_timeout. I created a dedicated timeout option.

plugins/modules/proxmox_ceph_pool.py
```diff
@@ -0,0 +1,260 @@
...
+        size=dict(type="int"),
+        timeout=dict(type="int", default=5),
+        target_size=dict(type="str"),
+        target_size_ratio=dict(type="int"),
+        state=dict(choices=["present", "absent"], required=True),
+    )
+
+
+def module_options():
+    return {}
```
IamLunchbox: You can omit this if you don't need it :)

plugins/modules/proxmox_ceph_pool.py
```diff
@@ -0,0 +1,260 @@
...
+                if task_success:
+                    self.module.exit_json(changed=True, msg=f"Ceph pool {name} deleted.")
+                else:
+                    self.module.fail_json(msg=f"Error occurred on task execution: {fail_reason}")
+        else:
+            self.module.exit_json(changed=False, msg=f"Ceph pool {name} not present.")
+
+
+def main():
+    module = create_proxmox_module(module_args(), **module_options())
```
IamLunchbox: You can omit `**module_options()` if you don't need it :)

plugins/module_utils/proxmox.py
```diff
@@ -418,3 +418,14 @@ def get_storage_content(self, node, storage, content=None, vmid=None):
             return self.proxmox_api.nodes(node).storage(storage).content().get(content=content, vmid=vmid)
         except Exception as e:
             self.module.fail_json(msg=f"Unable to list content on {node}, {storage} for {content} and {vmid}: {e}")
+
+    def check_node_on_cluster(self, node):
```
IamLunchbox:
```suggestion
    def node_in_cluster(self, node):
```

Sorry for the nitpick, but I think `in` is correct here. Within your module that does not matter too much, but since we share this functions between modules now... :)

plugins/module_utils/proxmox.py
```diff
@@ -418,3 +418,14 @@ def get_storage_content(self, node, storage, content=None, vmid=None):
...
+        """
+        Checks if the node is part of the cluster.
+
+        Args:
+            node(str): searched node.
+        """
+        nodes = self.proxmox_api.cluster.resources.get(type="node")
+        if node not in [item["node"] for item in nodes]:
+            self.module.fail_json(msg=f"Node {node} does not exist in the cluster.")
```
IamLunchbox: btw, there is also [this line](https://github.com/teslamania31/community.proxmox/blob/ce0116a7e3921dee71e210aaf3d0b9ebb30f140e/plugins/module_utils/proxmox.py#L227). Doesn't that do the same for your use case (fail if node does not exist)?

IamLunchbox (changes requested):

plugins/module_utils/proxmox.py
teslamania31: Yes, I just not need the node if it exists but the api endpoints seem similar

IamLunchbox: Ah now I see, the other function is only failing on an exception. 

I wonder if it easier to extend the given function with a `strict` parameter, which raises an error if the node does not exist? 

Maybe a bit more implicit, but saves a function definition nonetheless. What do you think is more concise?

IamLunchbox: @teslamania31

teslamania31: Sorry, I've been a bit busy.

So I removed the added shared function node_in_cluster and used the get_node shared function

plugins/modules/proxmox_ceph_pool.py
```diff
@@ -0,0 +1,256 @@
...
+        pools = self.proxmox_api.nodes(node).ceph.pool.get()
+        return any(pool["pool_name"] == name for pool in pools)
+
+    def get_pool(self, node, name):
+        return self.proxmox_api.nodes(node).ceph.pool(name).status.get()
+
+    def add_pool(self):
+        node = self.params["node"]
+        name = self.params["name"]
+        self.get_node(node)
```
IamLunchbox, line 189:
```suggestion
        self.get_node(node, strict=True)
```

Once #402 is merged this would fix the problem when get_node would return `None`.

Thulium-Drake: I'll merge that one first and have Github rebase this then :)

Thulium-Drake:
#402 has been merged!

Thulium-Drake:
NO CHANGELOG

Thulium-Drake:
Thanks for the module! :tada:

PR #342, opened 2026-03-21
style: fix ruff rule PLR0913

PendaGTP:
##### SUMMARY

Fixes the lint rule PLR0913 (19 errors): `Too many arguments in function definition (n > 5)`

I ignored most of these errors because fixing it would require more "deep" refactoring. I think it's preferable to just ignore them and address it when implementing new features, fixes, or more deeper refactoring module logic.  And also, this is more like a warnning, current code is still readable.

If you think that these errors should not be ignored but fixed, that's also ok, let me know.


PendaGTP:
Rebased

PR #349, opened 2026-03-22
Add `local` connection support.

### LINKED ISSUE #38: Support proxmoxer local mode
### Summary

Currently, Proxmox_KVM uses Proxmoxer in API mode. Proxmoxer also supports a SSH and (since Proxmoxer 1.3.0) local mode (the main difference being that local is intended for use when you're already SSH'd into the host, and SSH performs this SSH step itself). Supporting SSH mode is probably not worthwhile since it can't do anything API mode can't, but it would be handy to support 'local' mode of operation in Proxmoxer, since it is potentially fewer places to mess with credentials. Assuming that Ansible can already SSH into the Proxmox host, then using local mode, we don't need to also include / use API credentials in the playbook (either directly or indirectly), since the playbook would already be running on the host.

Notable changes for local mode would be that `api_host`, `api_user`, and `api_password` would be no longer required if using local mode. There would need to be a way to specify using `backend='local'` to Proxmox_KVM so that it can be passed to `ProxmoxAPI()` (which is called from `plugins/module_utils/proxmox.py` rather than directly in `plugins/modules/cloud/misc/proxmox_kvm.py` itself).

The only downside I'm aware of is that I believe Ansible would be essentially accessing Proxmox as `root` rather than some other limited user, so there may be cases where it is more desirable to do it the current way, such as providing a clear log within Proxmox that the action came from Ansible rather than the `root` user. However I would like to at least have the option to trade this reduction in logging clarity for simpler playbooks, since for some of them I need to run actions directly on the Proxmox host anyways (i.e. to run `qm importdisk ...` to import a cloudinit image to a VM disk, since this can't be done via the API), and it saves having to either store credentials in playbooks or pass them in somehow (via variables or vaults, etc).

I propose adding a `backend` option to `Proxmox_KVM`, which defaults to `https` (for normal API usage), but when set to `local` causes the `api_*` options to optional and/or ignored rather than required (`api_host` and `api_user` are normally required, the others optional) .

### Issue Type

Feature Idea

### Component Name

proxmox_kvm

### Additional Information

The first example in the docs, except using local mode (assumes that this playbook is being run on one of the nodes in the cluster - doesn't necessarily need to be the node the VM is created on):

```yaml
---
- name: Create VM using local mode
  hosts:
    - "{{ VM_HOST }}"

  tasks:
    - name: Create new VM with minimal options
      community.general.proxmox_kvm:
        backend: local
        name: spynal
        node: sabrewulf
```

vs the normal API mode as documented (could also be run on `"{{ VM_HOST }}"` or some other host, but probably don't want to run on `all`):
```yaml
---
- name: Create VM using API
  hosts: 
    - localhost

  tasks:
    - name: Create new VM with minimal options
      community.general.proxmox_kvm:
        api_user: root@pam
        api_password: secret
        api_host: helldorado
        name: spynal
        node: sabrewulf
```

### Code of Conduct

### LINKED ISSUE #348: Add `local` connection support.
##### SUMMARY

Add support for the `local` backend, allowing tasks to run directly on a Proxmox node.

My previous PR at https://github.com/ansible-collections/community.general/pull/4027 sadly didn't make it in before the collection was moved here.

Opening an issue since I don't have time to work on a new PR for this right away.

##### ISSUE TYPE

- Feature Idea

##### COMPONENT NAME

all

##### ADDITIONAL INFORMATION

Avoid the need to set up API keys when already running on a Proxmox node anyways.

reitermarkus:
##### SUMMARY

Fixes https://github.com/ansible-collections/community.proxmox/issues/348.
Fixes #38 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

all

##### ADDITIONAL INFORMATION

N/A


changelogs/fragments/999-local-backend.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+  - proxmox - use ``local`` backend if no ``api_host`` is specified (https://github.com/ansible-collections/community.proxmox/pull/999).
```
PendaGTP: Please update to reference this PR number instead of 999 (filename + link on changelog).

changelogs/fragments/349-local-backend.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+  - proxmox - use ``local`` backend if no ``api_host`` is specified (https://github.com/ansible-collections/community.proxmox/pull/349).
```
PendaGTP, line 5: Can you please add another changelog entries to indicate the change of defaults/required options.

Something in this idea:

```yml
- proxmox - set ``api_host`` as not required (to support ``local`` backend)
- proxmox - set ``root@pam`` as default for ``api_user```
```

Thulium-Drake:
Thanks a lot for this work! I'm pretty sure a lot of folks will like this! ;-)

Thulium-Drake (approved):

plugins/modules/proxmox_kvm.py
```diff
@@ -1300,25 +1301,24 @@ def create_vm(self, vmid, newid, node, name, memory, cpu, cores, sockets, update
...
+            if self.module.params["args"] is not None:
+                kwargs["args"] = self.module.params["args"]
+        else:
+            if self.module.params["args"] is not None:
+                self.module.fail_json(msg="args parameter require root@pam user. ")
 
-        if self.module.params["api_user"] != "root@pam" and self.module.params["skiplock"] is not None:
-            self.module.fail_json(msg="skiplock parameter require root@pam user. ")
+            if self.module.params["skiplock"] is not None:
+                self.module.fail_json(msg="skiplock parameter require root@pam user. ")
```
IamLunchbox, line 1313:
```suggestion
            if self.module.params["args"]:
                kwargs["args"] = self.module.params["args"]
        else:
            if self.module.params["args"]:
                self.module.fail_json(msg="Parameter `args` requires root@pam user. ")

            if self.module.params["skiplock"]:
                self.module.fail_json(msg="Parameter `skiplock` requires root@pam user. ")
```

plugins/module_utils/proxmox.py
```diff
@@ -32,9 +32,9 @@ def proxmox_auth_argument_spec():
         dict[str, dict]: Parameter names mapped to their configuration dictionaries.
     """
     return dict(
-        api_host=dict(type="str", required=True, fallback=(env_fallback, ["PROXMOX_HOST"])),
+        api_host=dict(type="str", fallback=(env_fallback, ["PROXMOX_HOST"])),
         api_port=dict(type="int", fallback=(env_fallback, ["PROXMOX_PORT"])),
-        api_user=dict(type="str", required=True, fallback=(env_fallback, ["PROXMOX_USER"])),
+        api_user=dict(type="str", default="root@pam", fallback=(env_fallback, ["PROXMOX_USER"])),
```
IamLunchbox, line 37: I don't think this is a sensible default. I'd keep the requirement to ask users for their username. @Thulium-Drake, what do you think?

Thulium-Drake: Well, for me it does make sense, Proxmox's installer does not create a useraccount by default (and I'm pretty sure 90%+ of the homelabs (re-)use `root@pam` as the main account).

And I think it definitely does make sense in the context where the local connection is used. And if this uses what I think it does (`pvesh`) that tool only works if you're root on the server. From the manpage:

```
DESCRIPTION
       The Proxmox VE management tool (pvesh) allows to directly invoke API function, without using the
       REST/HTTPS server.

           Note

           Only root is allowed to do that.
```

PendaGTP:
This will also fixes https://github.com/ansible-collections/community.proxmox/issues/38

IamLunchbox:
@reitermarkus Bump, if you are interested, you can fix the merge conflicts, the annotations and the sanity issue and we may merge soon :)

PendaGTP:
Hi @reitermarkus, following up here, remaining blockers are the merge conflicts, sanity check failures, and the review comments above. If you don't have time to finish this, happy to pick it up from your branch. Let us know either way.

PendaGTP:
Hi @reitermarkus, Since this PR has been open for few months I've picked up your branch and open #469 to continue the work. Thanks again for the initial work here!

Thulium-Drake:
Superseded by #469

PR #377, opened 2026-04-01
feat: add acme account module

PendaGTP:
##### SUMMARY

Related to: https://github.com/ansible-collections/community.proxmox/issues/376

This PR adds new modules:

- `proxmox_cluster_acme_account`
- `proxmox_cluster_acme_account_info`
- `proxmox_cluster_acme_accounts_info`

##### ISSUE TYPE

- Feature Pull Request

##### ADDITIONAL INFORMATION

Note: for the module `proxmox_cluster_acme_account`, during my tests the Proxmox API doesn't return list of emails in contact field on a get request, but this should be the case (based on PVE API docs), that's why I have added special code to handle this case.

EDIT: I'm not sure if we need to prefix by `_cluster` maybe `proxmox_acme_account` is suffisant, I have done this to match api spec, but maybe not the best for users, let me know


Thulium-Drake:
Looking at the rest of the modules, I'd say `proxmox_acme_*` should be clear enough. In the firewall mods there is a distinction between the cluster and non-cluster rules, so there it makes sense.

OTOH, having `_cluster` in it could also point users towards that it can only be used within a cluster, which isn't true AFAIK. So maybe the shorter name is better :-)

plugins/modules/proxmox_acme_account.py
```diff
@@ -0,0 +1,423 @@
...
+      - tos_url
+
+seealso:
+  - name: Certificate management (Proxmox documentation)
+    description: ACME accounts and certificates in Proxmox VE
+    link: https://pve.proxmox.com/pve-docs/pve-admin-guide.html#sysadmin_certificate_management
+  - module: community.proxmox.proxmox_acme_accounts_info
+    description: List ACME account names.
+  - module: community.proxmox.proxmox_acme_account_info
+    description: Retrieve information about a single ACME account.
```
Thulium-Drake, line 67: I think this is missing the PVE Authentication Guide?

plugins/modules/proxmox_acme_account_info.py
```diff
@@ -0,0 +1,137 @@
...
+  - community.proxmox.proxmox.actiongroup_proxmox
+  - community.proxmox.proxmox.documentation
+  - community.proxmox.attributes
+  - community.proxmox.attributes.info_module
+
+seealso:
+  - module: community.proxmox.proxmox_acme_accounts_info
+    description: List ACME account names.
+  - module: community.proxmox.proxmox_acme_account
+    description: Create, update or delete an ACME account.
```
Thulium-Drake, line 34: Same here

plugins/modules/proxmox_acme_account_info.py
```diff
@@ -0,0 +1,137 @@
...
+
+# Copyright (c) 2026, Clément Cruau (@PendaGTP) <38917281+PendaGTP@users.noreply.github.com>
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+
+DOCUMENTATION = r"""
+module: proxmox_acme_account_info
+short_description: Retrieve one ACME account
+version_added: "2.1.0"
```
Thulium-Drake: This can be `2.0.0`, it's not out yet ;-)
```suggestion
version_added: "2.0.0"
```

plugins/modules/proxmox_acme_account.py
```diff
@@ -0,0 +1,423 @@
...
+
+# Copyright (c) 2026, Clément Cruau (@PendaGTP) <38917281+PendaGTP@users.noreply.github.com>
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+
+DOCUMENTATION = r"""
+module: proxmox_acme_account
+short_description: ACME account management
+version_added: "2.1.0"
```
Thulium-Drake:
```suggestion
version_added: "2.0.0"
```

plugins/modules/proxmox_acme_accounts_info.py
```diff
@@ -0,0 +1,88 @@
...
+
+# Copyright (c) 2026, Clément Cruau (@PendaGTP) <38917281+PendaGTP@users.noreply.github.com>
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+
+DOCUMENTATION = r"""
+module: proxmox_acme_accounts_info
+short_description: List ACME account names
+version_added: "2.1.0"
```
Thulium-Drake:
```suggestion
version_added: "2.0.0"
```

plugins/modules/proxmox_acme_accounts_info.py
```diff
@@ -0,0 +1,88 @@
...
+  - community.proxmox.proxmox.actiongroup_proxmox
+  - community.proxmox.proxmox.documentation
+  - community.proxmox.attributes
+  - community.proxmox.attributes.info_module
+
+seealso:
+  - module: community.proxmox.proxmox_acme_account_info
+    description: Retrieve information about a single ACME account.
+  - module: community.proxmox.proxmox_acme_account
+    description: Create, update or delete an ACME account.
```
Thulium-Drake, line 27: And here :-)

Thulium-Drake (commented):
Some nitpickings, but I like how it's looking so far!

plugins/modules/proxmox_acme_account.py
PendaGTP: Are you talking about https://github.com/ansible-collections/community.proxmox/pull/383?

This will be added automaticaly since I added the ref to PVE auth on `plugins/doc_fragments/proxmox.py`

Or I miss something sorry..

PendaGTP:
Ready

plugins/modules/proxmox_acme_account.py
Thulium-Drake: ah, right, no, that should be fine then :-)

Thulium-Drake (approved):
🚀

PR #378, opened 2026-04-02
feat: add acme plugin dns module

PendaGTP:
##### SUMMARY

Related to:https://github.com/ansible-collections/community.proxmox/issues/376

Add new module `proxmox_cluster_acme_plugin_dns`

##### ISSUE TYPE

- Feature Pull Request

##### ADDITIONAL INFORMATION

I haven't tested the behavior of `validation-delay`, what will happen if this delay is greater than `api_timeout`? Should we set for this module `api_timeout` = `validation_delay`?

### DIFF
--- a/meta/runtime.yml
+++ b/meta/runtime.yml
@@ -12,6 +12,7 @@ action_groups:
     - proxmox_acme_account
     - proxmox_acme_account_info
     - proxmox_acme_accounts_info
+    - proxmox_acme_plugin_dns
     - proxmox_backup
     - proxmox_backup_info
     - proxmox_backup_schedule
@@ -20,7 +21,6 @@ action_groups:
     - proxmox_ceph_mon
     - proxmox_ceph_osd
     - proxmox_cluster
-    - proxmox_cluster_acme_plugin_dns
     - proxmox_cluster_firewall
     - proxmox_cluster_ha_resources
     - proxmox_cluster_ha_rules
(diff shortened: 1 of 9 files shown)


Thulium-Drake:
Regarding the timeouts, oof, good question, I don't know if the ACME API returns something along the lines of 'still working on it', I don't think so. In one of my roles I keep poking the ACME server every 60 seconds until it says the certificate has been issued.

But this does sound like an edge case that would be very valuable in the unit tests (and I hope to add the github.com/letsencrypt/pebble server in the molecule scenario at some point too :-) )

Thulium-Drake (approved):

PendaGTP:
I have added modules `proxmox_acme_plugin_info` and `proxmox_acme_plugins_info`

PendaGTP:
Ready!

Thulium-Drake (approved):
Nice work!

IamLunchbox:
Aw, not there is a merge conflict :(

PendaGTP:
Fixed!

IamLunchbox:
Ah in hindsight: Why two info modules? Both are very small? :thinking:

PendaGTP:
@IamLunchbox 

They don't provide same level of info, one can be use to list plugins and then do some processing (e.g compare/remove), and the other one to directly get info about a specific plugin. Each info module use different API.

I have done the same for acme account (list, get, manage).

This looks ok to me since these info modules doesn't comes with complexity and are easy to maintain.

PR #395, opened 2026-04-09
test: add a shared task to setup proxmox auth vars

PendaGTP:
##### SUMMARY

Simplify the configuration of Proxmox authentication for integration tests by adding a helper that sets env vars from a config file.

I haven't touched the existing integration tests yet, but I'm planning to. I was wondering how to avoid passing auth options on each task.

Not sure if this is the way to go. I'm quite new to this domain I found this pattern from the VMware collection.

Open to any ideas or thoughts, would really like to work on integration tests to make us more confident about changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,6 +59,10 @@ coverage.xml
 .hypothesis/
 .pytest_cache/
 
+# Integration test
+integration_config.yml
+inventory
+
 # Translations
 *.mo
 *.pot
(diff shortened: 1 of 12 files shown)


Thulium-Drake:
This is actually really nice, and less confusing then using the YAML anchors (https://docs.ansible.com/projects/ansible/latest/playbook_guide/playbooks_advanced_syntax.html#yaml-anchors-and-aliases-sharing-variable-values) I mean, they work, but they make it difficult to read (for me at least)

IamLunchbox (approved):
Good idea and nice work! This is way more concise ;)

PR #396, opened 2026-04-09
ci: check changelog fragments format

emmanueljamet:
##### SUMMARY
check changelog fragments format

The idea is to avoid/assist the manual check for fragments format

##### ISSUE TYPE
- Feature Pull Request (CI)

##### COMPONENT NAME
CI

##### ADDITIONAL INFORMATION
Feel free to add this check as required or not 🤷🏻

### DIFF
--- a/.github/workflows/changelog-format.yml
+++ b/.github/workflows/changelog-format.yml
@@ -4,7 +4,7 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 name: changelog fragments
-'on':
+on:
   pull_request:
     branches:
       - main
--- a/.github/workflows/changelog-format.yml
+++ b/.github/workflows/changelog-format.yml
@@ -1,0 +1,83 @@
+---
+# Copyright (c) 2026 Emmanuel Jamet <emmanueljamet@gmail.com>
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+name: changelog fragments
+'on':
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  changelog:
+    name: "Check changelog fragments"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+          sparse-checkout: |
+            changelogs
+
+      - name: Check fragments format
+        shell: python
+        env:
+          FRAGMENTS_PATH: "changelogs/fragments"
+        run : |
+          from os import getenv, listdir
+          from os.path import exists, isfile, join
+          from re import compile as recompile
+          from sys import exit as sysexit
+
+          from yaml import safe_load
+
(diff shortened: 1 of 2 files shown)


.github/workflows/changelog-format.yml
```diff
@@ -0,0 +1,77 @@
...
+                              "removed_features",
+                              "security_fixes",
+                              "bugfixes",
+                              "known_issues",
+                              "trivial",
+                          ]:
+                              print(f"unrecognized change type: {typ}")
+                              sysexit(1)
+
+                          regex = recompile(r"\w+ - [a-z].*\.")
```
IamLunchbox: This won't match the example in the ansible docs `bitbucket_* modules - client_id is no longer marked as ``no_log=true``. If you relied on its value not showing up in logs and output, mark the whole task with ``no_log: true`` (https://github.com/ansible-collections/community.general/pull/2045).` though. But otherwise a good idea!

IamLunchbox (commented):
Thank you for this! good idea. Did you want to include the changelog fragment to proof, that your ci works? Because it is not actually needed per the conventions ;)

.github/workflows/changelog-format.yml
emmanueljamet: good catch!

fixed.

Thulium-Drake:
Awesome, this is a great idea! I like it :-)

Not sure on how we can make it mandatory though, but if it will fail a (part) of the pipeline, it will show up anyway. And when we merge code that does not pass CI, we often have a reason for it. Like with a lot of the refactoring that was done in the past few weeks.

Thulium-Drake (approved):

.github/workflows/changelog-format.yml
```diff
@@ -0,0 +1,77 @@
+---
+# Copyright (c) 2026 Emmanuel Jamet <emmanueljamet@gmail.com>
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+name: changelog fragments
+'on':
```
IamLunchbox:
```suggestion
on:
```

IamLunchbox (changes requested):
I have another problem: this will always fail for everybody. Since the PR number is not known beforehand, most of the time PRs get created without fragment. 

And this will fail then. And not be run again, because the condition is only met on opening the PR. Therefore I think it is good if we think this through how the run conditions could be improved :)

emmanueljamet:
> I have another problem: this will always fail for everybody. Since the PR number is not known beforehand, most of the time PRs get created without fragment.
> 
> And this will fail then. And not be run again, because the condition is only met on opening the PR. Therefore I think it is good if we think this through how the run conditions could be improved :)

It won't, as by default this kind of events trigger the run: "By default, a workflow only runs when a pull_request event's activity type is `opened`, `synchronize`, or `reopened`"

(https://docs.github.com/fr/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request)

As you can see with each of my commits on this PR

PR #402, opened 2026-04-16
Add strict mode to get_node

IamLunchbox:
This PR adds a strict mode get_node to create a shared function, which enables early failures if a required node is needed but does not exist.

### DIFF
--- a/plugins/module_utils/proxmox.py
+++ b/plugins/module_utils/proxmox.py
@@ -235,12 +235,13 @@ class ProxmoxAnsible:
         except Exception as e:
             self.module.fail_json(msg=f"Unable to retrieve Proxmox VE version: {e}")
 
-    def get_node(self, node):
+    def get_node(self, node, strict=False):
         """
         Filters all known PVE nodes for the given node name.
 
         Args:
             node(str): The name of the node.
+            strict(bool): Fail if the node does not exist.
 
         Returns:
             dict | None: The node information provided by the api path GET /nodes.
@@ -249,6 +250,8 @@ class ProxmoxAnsible:
             nodes = [n for n in self.proxmox_api.nodes.get() if n["node"] == node]
         except Exception as e:
             self.module.fail_json(msg=f"Unable to retrieve Proxmox VE node: {e}")
+        if not nodes and strict:
+            self.module.fail_json(msg=f"Node {node} does not exist")
         return nodes[0] if nodes else None
 
     def get_nextvmid(self):


IamLunchbox:
@emmanueljamet CI shoudlve failed here, I guess your script is not working as intended. I think because main contains all changelog fragments until a relase happens. Therefore, this ci fails only on the first PR after a release :(

I think we should use https://docs.github.com/en/rest/pulls/pulls?apiVersion=2026-03-10#list-pull-requests-files to fix this issue. I'll draft a PR.

emmanueljamet:
> @emmanueljamet CI shoudlve failed here, I guess your script is not working as intended. I think because main contains all changelog fragments until a relase happens. Therefore, this ci fails only on the first PR after a release :(
> 
> I think we should use https://docs.github.com/en/rest/pulls/pulls?apiVersion=2026-03-10#list-pull-requests-files to fix this issue. I'll draft a PR.

Why should it fail on this PR?

The check only checks the format of existing fragments, not if you need to add one.

Unless we have a very precise rule, it will be hard to avoid false positives if we check on the PR files.

Thulium-Drake:
@IamLunchbox #403 Seems to work just fine, do you want to use this PR to also check what happens if you put in a faulty changelog format?

Thulium-Drake:
@IamLunchbox does this also fix #379 ?

IamLunchbox:
> > @emmanueljamet CI shoudlve failed here, I guess your script is not working as intended. I think because main contains all changelog fragments until a relase happens. Therefore, this ci fails only on the first PR after a release :(
> > 
> > I think we should use https://docs.github.com/en/rest/pulls/pulls?apiVersion=2026-03-10#list-pull-requests-files to fix this issue. I'll draft a PR.
> 
> Why should it fail on this PR?
> 
> The check only checks the format of existing fragments, not if you need to add one.
> 
> Unless we have a very precise rule, it will be hard to avoid false positives if we check on the PR files.

Ah, I misunderstood your intent then. I assumed you wanted to cover reminding people of fragments as well :)

Well, now it fails more often, but it does both now. 

And Thulium and I can disable it with NO CHANGELOG

IamLunchbox:
> @IamLunchbox #403 Seems to work just fine, do you want to use this PR to also check what happens if you put in a faulty changelog format?

But we don't need one in this PR :( I will make sure to add one the next time :)

IamLunchbox:
> @IamLunchbox does this also fix #379 ?

I think this issue is not caused by community.proxmox. neither does it makes sense code wise that it fails, nor was it reproducible. 🤷‍♂️

PR #406, opened 2026-04-20
Feat/add acme certificate

PendaGTP:
##### SUMMARY

Add new modules to manage acme certificates.

Related to: #376

##### ISSUE TYPE

- Feature Pull Request

##### ADDITIONAL INFORMATION

~~On the module documentation I reference the module acme_plugin, so merging the other PR before is preferable.~~

### DIFF
--- a/plugins/module_utils/proxmox_acme_certificate.py
+++ b/plugins/module_utils/proxmox_acme_certificate.py
@@ -1,0 +1,75 @@
+#
+# Copyright (c) 2026, Clément Cruau (@PendaGTP) <38917281+PendaGTP@users.noreply.github.com>
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+"""Shared helpers for Proxmox ACME certificate modules."""
+
+_ACME_DOMAIN_SLOTS = 6  # acmedomain0 .. acmedomain5
+
+
+def parse_acme_config(node_config):
+    """Parse a node config dict into structured ACME settings.
+
+    Returns a dict with ``account`` (str or None) and ``domains`` (list of dicts
+    with keys ``domain``, ``plugin``, ``alias``).
+    """
+    account = None
+    domains = []
+
+    acme_raw = node_config.get("acme")
+    if acme_raw:
+        for part in acme_raw.split(","):
+            kv = part.strip().split("=", 1)
+            if len(kv) == 2 and kv[0].strip() == "account":  # noqa: PLR2004
+                account = kv[1].strip()
+
+    for i in range(_ACME_DOMAIN_SLOTS):
+        key = f"acmedomain{i}"
+        raw = node_config.get(key)
+        if not raw:
+            continue
+        domain_entry = _parse_acmedomain_string(raw)
+        if domain_entry:
+            domains.append(domain_entry)
+
+    return {"account": account, "domains": domains}
+
+
+def _parse_acmedomain_string(value):
+    """Parse a single ``acmedomainN`` property string into a dict."""
+    result = {"domain": None, "plugin": None, "alias": None}
+    for part in value.split(","):
+        kv = part.strip().split("=", 1)
+        if len(kv) != 2:  # noqa: PLR2004
+            continue
+        key, val = kv[0].strip(), kv[1].strip()
+        if key in result:
+            result[key] = val
+    return result if result["domain"] else None
+
+
+def cert_info_to_ansible_result(cert):
+    """Transform a single certificate info dict from the Proxmox API into Ansible return values."""
+    san = cert.get("san") or []
+    if isinstance(san, str):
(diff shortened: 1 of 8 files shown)


Thulium-Drake:
Which other modules do you mean? :) #378 and #377 have been merged already ;-)

PendaGTP:
Oh yes sorry, perfect!

plugins/module_utils/proxmox_acme_certificate.py
```diff
@@ -0,0 +1,75 @@
...
+
+def cert_info_to_ansible_result(cert):
+    """Transform a single certificate info dict from the Proxmox API into Ansible return values."""
+    san = cert.get("san") or []
+    if isinstance(san, str):
+        san = [s.strip() for s in san.split(",") if s.strip()]
+
+    fingerprint = cert.get("fingerprint") or ""
+    not_before = cert.get("notbefore") or ""
+    not_after = cert.get("notafter") or ""
```
IamLunchbox:
```suggestion
    fingerprint = cert.get("fingerprint", "")
    not_before = cert.get("notbefore", "")
    not_after = cert.get("notafter", "")
```

plugins/module_utils/proxmox_acme_certificate.py
```diff
@@ -0,0 +1,75 @@
...
+
+    if isinstance(not_before, (int, float)):
+        not_before = str(int(not_before))
+    if isinstance(not_after, (int, float)):
+        not_after = str(int(not_after))
```
IamLunchbox: Why do you do str(int()) when you return int() of those objects again?

plugins/modules/proxmox_acme_account.py
```diff
@@ -369,22 +370,15 @@ def _account_endpoint(self, name=None):
         return base if not name else base(name)
 
     def _wait_acme_task(self, taskid):
-        node = self._node_from_upid(taskid)
-        ok, err = self.api_task_complete(node, taskid, 30)
+        node = self.upid_to_node(taskid)
+        ok, err = self.api_task_complete(node, taskid, ACME_TASK_TIMEOUT)
```
IamLunchbox, line 374: Does it make sense to extend  `api_task_ok` with a call to `api_task_complete`, e.g. with a feature flag? This pattern here probably happens a lot in our modules: You wait for completion and then check if it was ok to proceed. If everyone needs to do it, maybe we should provide a way to do it in a builtin way.

plugins/modules/proxmox_acme_certificates_info.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+def module_args():
+    return dict(
+        node_name=dict(type="str", required=True),
+    )
+
+
+def module_options():
+    return {}
```
IamLunchbox, line 108: The way you wrote `create_proxmox_module` makes it possible to leave this out ;)

plugins/modules/proxmox_acme_certificate.py
```diff
@@ -0,0 +1,496 @@
...
+CERTIFICATE_TASK_TIMEOUT = 120
+ACME_DOMAIN_SLOTS = 6  # acmedomain0 .. acmedomain5
+
+
+def build_acme_property_string(account):
+    """Build the ``acme`` node config property string.
+
+    Example output: ``"account=production"``
+    """
+    return f"account={account}"
```
IamLunchbox: This function seems a bit short to be a function at all?

plugins/modules/proxmox_acme_certificate.py
```diff
@@ -0,0 +1,496 @@
...
+    ACME and custom certificates are stored as ``pveproxy-ssl.pem``,
+    while the self-signed cert is ``pve-ssl.pem``.
+
+    Returns the matching certificate dict, or None.
+    """
+    if not certificates:
+        return None
+
+    for cert in certificates:
+        filename = cert.get("filename") or ""
```
IamLunchbox:
```suggestion
        filename = cert.get("filename", "")
```

IamLunchbox (approved):

plugins/modules/proxmox_acme_certificates_info.py
PendaGTP: Yes but this is the pattern that have already been applied on other modules, e.g `proxmox_role`, `proxmox_pool`, `proxmox_cluster_firewall`, etc.

Maybe can we let it like this or plan a refactor for all other module which doesn't has module options.

plugins/modules/proxmox_acme_account.py
PendaGTP: Yes I see what you mean, if it's ok for you I perfer do this refactor in an other PR dedicated to PVE task waiting handling.

PendaGTP:
Rebased, patchs applied.

Thulium-Drake:
NO CHANGELOG

PendaGTP:
I miss something?

Thulium-Drake:
@PendaGTP no, I just had to run CI again to have it pass the changelog check ;-) it's done now :)

Thank you for your work! :tada:

PR #412, opened 2026-04-22
fix(pool): pool member retrieval

PendaGTP:
##### SUMMARY

Currently modules `proxmox_pool` and `proxmox_pool_member` are failing when checking pool member because `get_pool` return a list instead of an dict.

This is caused by `?poolid=poolid` which return a list but we expect only one pool returned since poolid are unique, we need to keep `?poolid=poolid` to support sub pool e.g `test/example`.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

`proxmox_pool` 
`proxmox_pool_member`


IamLunchbox:
Thanks!

PR #415, opened 2026-04-23
proxmox: fix LXC tags always reported as changed

cesasol:
## Summary
- `tags` on LXC containers is accepted as a list by the module, but Proxmox stores it as a semicolon-delimited string. The existing diff fell through to `str(value) != str(current_config[arg])`, comparing `\"['foo', 'bar']\"` against `'foo;bar'` — always different, so updates always reported `changed=True`.
- Introduces a small `_LIST_FIELDS` table mapping list-valued module args to the PVE separator and compares them as sets, so the diff is accurate and order-insensitive (matching PVE's server-side sort of tags).
- Includes a changelog fragment.

## Test plan
- [x] `nox -Re codeqa` (ruff check) — passes.
- [x] `nox -Re formatters` (ruff format) — no changes.
- [x] `nox -Re ansible-test-units-devel -- --python 3.13 tests/unit/plugins/modules/test_proxmox_lxc.py` — passes.
- [x] `nox -Re ansible-test-sanity-devel` — passes.
- [ ] Manual: update an LXC container twice with the same `tags` list and confirm the second run is `changed=False`.
- [ ] Manual: update with a reordered `tags` list and confirm `changed=False`.
- [ ] Manual: update with an added or removed tag and confirm `changed=True`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/changelogs/fragments/413-lxc-tags-diff.yml
+++ b/changelogs/fragments/413-lxc-tags-diff.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - proxmox - fix ``tags`` always being reported as changed on LXC container updates because the list value was compared against Proxmox's semicolon-delimited string form (https://github.com/ansible-collections/community.proxmox/pull/413).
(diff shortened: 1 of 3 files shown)


IamLunchbox (approved):

Thulium-Drake:
Thanks for this! :rocket:

PR #419, opened 2026-04-30
fix(acme_account): set no_log on sensitive value eab_kid

PendaGTP:
##### SUMMARY

The field `eab_kid` defines an external account binding which is a sensitive value and should be marked as `no_log`.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

`proxmox_acme_account`


Thulium-Drake:
Thanks for this :-)

PR #422, opened 2026-05-08
(proxmox,proxmox_kvm) Add destroy_unreferenced_disks parameter

DBS-ST-VIT:
##### SUMMARY
The Proxmox API supports removing unreferenced disks (i.e. disks that contain the VMID in their name but are not referenced in the container/VM config via the destroy-unreferenced-disks parameter. This PR adds this option as a new module parameter "destroy_unreferenced_disks" to both proxmox and proxmox_kvm modules.

The module parameter defaults to 'False' and the modules themselves send '1' via the API if it is set to True.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
proxmox
proxmox_kvm

##### ADDITIONAL INFORMATION
Since Proxmox usually takes good care of making sure there are no unreferenced volumes, the easiest way to test this is to
- create a VM/Container with an additional volume besides the root volume attached
- stop the VM/Container and manually remove the reference to that volume from the config file

### DIFF
--- a/changelogs/fragments/422-destroy-unref-disks-option.yml
+++ b/changelogs/fragments/422-destroy-unref-disks-option.yml
@@ -1,0 +1,4 @@
+---
+minor_changes:
+  - proxmox - add ``destroy_unreferenced_disks`` parameter (https://github.com/ansible-collections/community.proxmox/pull/422).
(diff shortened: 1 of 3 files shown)


IamLunchbox (approved):

Thulium-Drake:
Thank you for this! :rocket: :grin:

DBS-ST-VIT:
You're welcome :)

Issue #300, opened 2026-02-26
Proposal: Split `proxmox_firewall` and new firewall features


PendaGTP:
## Summary

Split `proxmox_firewall` into four sub-modules and introduce three new module to manage new firewall options (cluster-level node-level and vm/lxc-level)

### Context

Currently, the module `proxmox_firewall` manages "lot of things", which makes it difficult to maintain and, in my opinion, impractical for users.
I propose to split it into smaller module and focusing on one feature at a time.

Current the module handle (each one will be new dedicated module):

- Security Group
- Rules
- Alias
- IP Sets

The module doesn't handle (each one will be new dedicated module):

- Firewall options on node-level and VM/container-level
- Firewall options on cluster-level (#297)

I think we can work on these modules without removing "proxmox_firewall," but gradually indicate in the module documentation the recommendation to use a dedicated module, then deprecate the "proxmox_firewall" module.

I don't think all modules should comes on a same time, they can be added one by one, without impacting users.

Bellow modules that I have in mind that could fit well. For each module I can draft module documention if this can helps use define module interface.

As mentionned this is a proposal, a starting point for thinking about the path we want to take for managing firewall features. Please feel free to share your thoughts.

Related issues (just for tracking):

- #205
- #278

#### `proxmox_firewall_rules`

Description: Manages cluster-level, node-level or VM/container-level firewall rules
API:

- `/api2/json/cluster/firewall/rules`
- `/api2/json/nodes/{node}/firewall/rules`
- `/api2/json/nodes/{node}/lxc/{vmid}/firewall/rules`
- `/api2/json/nodes/{node}/qemu/{vmid}/firewall/rules`

#### `proxmox_cluster_firewall_security_group`

Description: Manages security group defined at cluster level
API:

- `/api2/json/cluster/firewall/groups/`

#### `proxmox_firewall_options`

Description: Manages firewall options on VM/container-level
API:

- `/api2/json/nodes/{node}/qemu/{vmid}/firewall/options`
- `/api2/json/nodes/{node}/lxc/{vmid}/firewall/options`

#### `proxmox_node_firewall_options`

Description: Manages firewall options on node-level
API:

- `/api2/json/nodes/{node}/firewall/options`

#### `proxmox_firewall_ipset`

Description: Manage IPSET at cluster-level, VM/container-level
API:

- `/api2/json/cluster/firewall/ipset`
- `/api2/json/nodes/{node}/qemu/{vmid}/firewall/ipset`
[...]

PR #425, opened 2026-05-11
Feat/add cluster firewall security group

PendaGTP:
##### SUMMARY

Add new module `proxmox_cluster_firewall_security_group`, dedicated module for cluster security groups.

Related to: https://github.com/ansible-collections/community.proxmox/issues/300

This extraction from `proxmox_firewall` comes with key beneficts:

- supports full check mode
- provides a clear declarative rules model:
  - `rules` omitted = keep existing rules untouched
  - `rules: []` = remove all rules
  - `rules: [...]` = reconcile full ordered ruleset (list index = rule position

Some "side" beneficts:

- support updating the security group comment
- support deleting a security group with existing rules by removing them first
- no "pos" required in rules
- retry logic to handle put concurrency issue (tested and can occur when patching large amount of rules)

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

`proxmox_cluster_firewall_security_group`

##### ADDITIONAL INFORMATION

Unit and integration tests were generated with AI and reviewed/adjusted manually.

### DIFF
--- a/plugins/modules/proxmox_cluster_firewall_security_group.py
+++ b/plugins/modules/proxmox_cluster_firewall_security_group.py
@@ -267,13 +267,13 @@ def _build_create_rule_payload(desired_rule, position, group_name):
         "enable": ansible_to_proxmox_bool(desired_rule.get("enabled", True)),
     }
     for ansible_key, api_key in _OPTIONAL_RULE_TO_API.items():
-        if desired_rule.get(ansible_key) is not None:
+        if desired_rule.get(ansible_key):
             payload[api_key] = desired_rule[ansible_key]
-    if group_name is not None:
+    if group_name:
         payload["group"] = group_name
-    if position is not None:
+    if position:
         payload["pos"] = position
-    return {k: v for k, v in payload.items() if v is not None or k == "enable"}
+    return {k: v for k, v in payload.items() if v or k == "enable"}
 
 
 def _build_update_rule_payload(desired_rule, current_rule):
--- a/plugins/modules/proxmox_cluster_firewall_security_group.py
+++ b/plugins/modules/proxmox_cluster_firewall_security_group.py
@@ -1,0 +1,660 @@
+#!/usr/bin/python
+
+# Copyright (c) 2026, Clément Cruau (@PendaGTP) <38917281+PendaGTP@users.noreply.github.com>
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+DOCUMENTATION = r"""
+module: proxmox_cluster_firewall_security_group
+short_description: Cluster firewall security group management for Proxmox VE
+version_added: "2.0.0"
+author:
+  - Clément Cruau (@PendaGTP)
+description:
+  - Manage cluster-level firewall security groups in Proxmox VE.
+  - Security groups hold ordered firewall rules that can be attached to guest rules.
+  - If O(rules) is omitted when O(state=present), the module only creates the group and/or
+    updates its comment; existing rules are not read or modified.
+  - If O(rules) is set (including an empty list), the module synchronizes the full ordered
+    ruleset, mapping list index 0 to position 0, etc.
+attributes:
+  check_mode:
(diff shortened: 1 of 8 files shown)


Thulium-Drake:
NO CHANGELOG

Thulium-Drake (approved):
PVE Firewall is still a bit magic to me, but the code looks solid. I've looked through it a few times now and I don't see any weird things :-)

🚀

plugins/modules/proxmox_cluster_firewall_security_group.py
```diff
@@ -0,0 +1,683 @@
+#!/usr/bin/python
+
+# Copyright (c) 2026, Clément Cruau (@PendaGTP) <38917281+PendaGTP@users.noreply.github.com>
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+DOCUMENTATION = r"""
+module: proxmox_cluster_firewall_security_group
+short_description: Cluster firewall security group management for Proxmox VE
+version_added: "2.0.0"
```
IamLunchbox:
```suggestion
version_added: "2.1.0"
```

plugins/modules/proxmox_cluster_firewall_security_group.py
```diff
@@ -0,0 +1,683 @@
...
+        "enable": ansible_to_proxmox_bool(desired_rule.get("enabled", True)),
+    }
+    for ansible_key, api_key in _OPTIONAL_RULE_TO_API.items():
+        if desired_rule.get(ansible_key) is not None:
+            payload[api_key] = desired_rule[ansible_key]
+    if group_name is not None:
+        payload["group"] = group_name
+    if position is not None:
+        payload["pos"] = position
+    return {k: v for k, v in payload.items() if v is not None or k == "enable"}
```
IamLunchbox:
```suggestion
        if desired_rule.get(ansible_key):
            payload[api_key] = desired_rule[ansible_key]
    if group_name:
        payload["group"] = group_name
    if position:
        payload["pos"] = position
    return {k: v for k, v in payload.items() if v or k == "enable"}
```

I think this was anothe reoccuring pattern I think. You commented, that you use an llm to help the coding - would you mind reminding it in the prompt to omit this pattern? :)

`if position` does already apply `bool(position)` implicitly.

IamLunchbox (approved):

IamLunchbox:
Just my two cents, since we are already at 2.0.0 we have to adjust the version though :innocent: 

Thank for for handling this! :rocket:

plugins/modules/proxmox_cluster_firewall_security_group.py
PendaGTP: Just a general question, why do you we not bump minor/path version for this collection? Thanks!

plugins/modules/proxmox_cluster_firewall_security_group.py
PendaGTP: Thanks for your comment. To be honest, I'm new to Python and I not always sure what the best way to do this thing, so this clarification really helps!

plugins/modules/proxmox_cluster_firewall_security_group.py
IamLunchbox: What do you mean exactly? That we develop in a dedicated release branch?

PendaGTP: I mean I more used to bump minor/patch version on each change and release a new version each time, but we target each time 2.0.0 (and before 1.6.0). Anyway the current workflow is totaly ok to me. Sorry if I'm not really clear.. :/

Thulium-Drake: So we (try) to keep a semantic versioning scheme (https://semver.org/) which means that any _new_ functionality is a new minor version. And as `2.0.0` has been released, the next one will be `2.1.0` :-)

And all bugfixes etc. (should we have enough to fill up a release or important enough to release on it's own) can warrant a patch release, e.g. `2.0.1`.

Thulium-Drake:
Thanks for making this functionality :rocket: !

PR #438, opened 2026-05-24
feat(domain): add openid audiences parameter

PendaGTP:
##### SUMMARY

Add parameter `openid_audiences` for openid realm.

This parameter has been added in Proxmox 9.2: https://bugzilla.proxmox.com/show_bug.cgi?id=5076

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

`proxmox_domain`

### DIFF
--- a/changelogs/fragments/437-add-openid-audiences-parameter.yaml
+++ b/changelogs/fragments/437-add-openid-audiences-parameter.yaml
@@ -1,3 +1,3 @@
 ---
 minor_changes:
-  - proxmox_domain - Add ``openid_audiences`` parameter introduced in Proxmox VE 9.2 (https://github.com/ansible-collections/community.proxmox/pull/437).
+  - proxmox_domain - add ``openid_audiences`` parameter introduced in Proxmox VE 9.2 (https://github.com/ansible-collections/community.proxmox/pull/437).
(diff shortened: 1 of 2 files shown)


IamLunchbox:
If it is only applicable for 9.2+, maybe we need a check to not send this parameter to a 9.1 cluster? Or we add it to the parameter docs?

IamLunchbox:
@PendaGTP bump regarding my question

PendaGTP:
I don't know what we should do, my initial thinking was: this is a minor version bump, the api should return an explicit error on 9.1 (haven't tested yet), adding minor check will add more maintenance burden over time.

But I might be wrong and will be happy to go whichever way

Thulium-Drake:
If the API reports a clear error for the user that is actionable, I think it's OK to leave it as is. We don't want to replicate the argument validation for the entire API :-)

PendaGTP:
I have just tested on `9.1.9` and the error reported is: `Task failed: Module failed: 400 Bad Request: Parameter verification failed. - {'audiences': 'property is not defined in schema and the schema does not allow additional properties`

To me this error is explicit and so no need to add any warning or note in the module docs. What do you think?

Thulium-Drake:
Yes, that error is clear enough for me, and I don't think it will be added to 9.1, so it's LGTM

Thulium-Drake (approved):

PR #445, opened 2026-05-28
feat(proxmox_user): add update password option

### LINKED ISSUE #442: Editing proxmox user using api key
##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

community.proxmox.proxmox_user module
##### ANSIBLE VERSION

```paste below
ansible [core 2.20.5]
  config file = ~/project/ansible.cfg
  configured module search path = ['~/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = ~/.pyenv/versions/ansible/lib/python3.14/site-packages/ansible
  ansible collection location = ~/.ansible/collections:/usr/share/ansible/collections
  executable location = ~/.pyenv/versions/ansible/bin/ansible
  python version = 3.14.4 (main, Apr  8 2026, 04:02:31) [GCC 15.2.0] (~/.pyenv/versions/ansible/bin/python3)
  jinja version = 3.1.6
  pyyaml version = 6.0.3 (with libyaml v0.2.5)
```

##### COLLECTION VERSION

```paste below
 ~/.ansible/collections/ansible_collections
Collection        Version
----------------- -------
community.proxmox 2.0.0
```

##### CONFIGURATION

```paste below
CALLBACKS_ENABLED(~/project/ansible.cfg) = ['ansible.posix.profile_tasks', 'ansible.posix.profile_roles']
CONFIG_FILE() = ~/project/ansible.cfg
DEFAULT_FILTER_PLUGIN_PATH(~/project/ansible.cfg) = ['~/project/filter_plugins']
DEFAULT_HASH_BEHAVIOUR(~/project/ansible.cfg) = merge
DEFAULT_HOST_LIST(~/project/ansible.cfg) = ['~/project/hosts.yml']
DEFAULT_LOAD_CALLBACK_PLUGINS(~/project/ansible.cfg) = True
DEFAULT_ROLES_PATH(~/project/ansible.cfg) = ['~/project/roles', '~/project/roles/init']
DEFAULT_STDOUT_CALLBACK(~/project/ansible.cfg) = default
DEFAULT_VAULT_PASSWORD_FILE(~/project/ansible.cfg) = ~/.ansible-vault
EDITOR(env: EDITOR) = vim
INTERPRETER_PYTHON(~/project/ansible.cfg) = auto_silent
RETRY_FILES_ENABLED(~/project/ansible.cfg) = False
```

##### OS / ENVIRONMENT

Ansible client -> Ubuntu 26.04 LTS
Target -> Proxmox VE 9.1.7
API token "user_provisioning" on root Proxmox user with `PVEUserAdmin` role on `/`

##### STEPS TO REPRODUCE

Creating a user with the proxmox_user module and running the playbook a second time after the user has been created.

```yaml
---
- name: Proxmox_users
  hosts: proxmox
  tasks:
    - name: "Proxmox_users | Manage user"
      community.proxmox.proxmox_user:
        api_user:         "root@pam"
        api_host:         "{{ proxmox_users_api_host }}"
        api_token_id:     "{{ proxmox_users_api_token_id }}"
        api_token_secret: "{{ proxmox_users_api_token_secret }}"
        name:             "test@pve"
        firstname:        "test"
        lastname:         "user"
        password:         "{{ proxmox_users_password }}"
        comment:          "Managed by Ansible"
        enable:           true
        validate_certs:   false
      delegate_to: localhost
```

##### EXPECTED RESULTS

According to the documentation on the password parameter:
> Initial password.
> Only for PVE Authentication Realm users.
> Parameter is ignored when user already exists or `state=absent`.

So I would expect the password to be ignored by the module if the user already exists, and the module not to try accessing it.
##### ACTUAL RESULTS

On the second run from the playbook (running when the user exists), it raises a permission error.

```paste below
TASK [Proxmox_users | Manage user] *********************************************
[ERROR]: Task failed: Module failed: Failed to update user password for user ID test@pve: 403 Forbidden: Permission check failed (URI '/access/password' not available with API token, need proper ticket.
Origin: ~/project/proxmox_users.yml:5:7

3   hosts: proxmox
4   tasks:
5     - name: "Proxmox_users | Manage user"
         ^ column 7

task path: ~/project/proxmox_users.yml:5
fatal: [proxmox -> localhost]: FAILED! =>
    changed: false
    msg: 'Failed to update user password for user ID test@pve: 403 Forbidden:
        Permission check failed (URI ''/access/password'' not available with API token,
        need proper ticket.'
    userid: test@pve
```

PendaGTP:
##### SUMMARY

Add an option `update_password` on `proxmox_user` to indicate if the password should be updated if the user already exists.

Fixes #442

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

`proxmox_user`

### DIFF
--- a/plugins/modules/proxmox_user.py
+++ b/plugins/modules/proxmox_user.py
@@ -391,7 +391,7 @@ class ProxmoxUserAnsible(ProxmoxAnsible):
                     )
 
             # We have no way of testing if the user's password needs to be changed
-            update_password = self.module.params.get("update_password") or True
+            update_password = self.module.params.get("update_password")
             if password and update_password:
                 try:
                     self.proxmox_api.access.password.put(userid=userid, password=password)
--- a/plugins/modules/proxmox_user.py
+++ b/plugins/modules/proxmox_user.py
@@ -74,10 +74,14 @@ options:
     type: str
   password:
     description:
-      - Initial password.
+      - Password for the user.
       - Only for PVE Authentication Realm users.
-      - Parameter is ignored when user already exists or O(state=absent).
     type: str
+  update_password:
+    description:
+      - Whether to update the password of the user.
+    type: bool
+    default: true
   tokens:
     description: List of API tokens associated to the user.
     type: list
@@ -171,6 +175,7 @@ def module_args():
         lastname=dict(type="str"),
         keys=dict(type="str", no_log=True),
         password=dict(type="str", no_log=True),
+        update_password=dict(type="bool", default=True),
         tokens=dict(
             type="list",
             no_log=False,
@@ -386,8 +391,8 @@ class ProxmoxUserAnsible(ProxmoxAnsible):
                     )
 
             # We have no way of testing if the user's password needs to be changed
-            # so, if it's provided we will update it anyway
-            if password:
+            update_password = self.module.params.get("update_password") or True
+            if password and update_password:
                 try:
                     self.proxmox_api.access.password.put(userid=userid, password=password)
                     self.module.exit_json(changed=True, userid=userid, msg=f"User {userid} updated")
--- a/changelogs/fragments/441-proxmox-user-update-password.yml
+++ b/changelogs/fragments/441-proxmox-user-update-password.yml
@@ -1,0 +1,3 @@
+---
+minor_changes:
+  - proxmox_user - add ``update_password`` parameter to choose whether to update the password of the user (https://github.com/ansible-collections/community.proxmox/pull/441).


IamLunchbox:
Thanks for taking the time to fix that!

Do you have an idea why #442 had a 403 error code as a result? Shouldnt a user being able to create an account be also able to set the password?

What do you think to always apply the mentioned password instead of adding this option? :thinking: `password` itself seems to be optional already.

plugins/modules/proxmox_user.py
```diff
@@ -386,8 +391,8 @@ def create_update_user(  # noqa: PLR0913
                     )
 
             # We have no way of testing if the user's password needs to be changed
-            # so, if it's provided we will update it anyway
-            if password:
+            update_password = self.module.params.get("update_password") or True
```
IamLunchbox: This pattern does not make sense. 

```suggestion
            update_password = self.module.params.get("update_password")
```


This is enough and more pythonic - but it could be left out as well, since Ansible will set the default.

Could you maybe keep a lookout in the future, that this pattern is not repeated any more? 😄  I think I mentioned that some times in the past already :s

IamLunchbox (changes requested):

PendaGTP:
Based on my test, you can't update user password using api token auth even with admin privileges.

But as you mention you can just leave password empty when using api token and this will works (you can update other fields). 

So maybe this fixes is maybe not really desired, user just have to use user auth for password set/update and when using api token, remove password field.

I will ask more info in the issue, for now I put in draft this PR.


edit: I think we may just need to update the documentation and remove: `Parameter is ignored when user already exists or state=absent.`

plugins/modules/proxmox_user.py
PendaGTP: Sorry, in fact this is not the first time, you already help me a lot on this.. I will be more focus on this next times!

PendaGTP:
Replaced by #451 which address the "root" issue (update user password using api auth token)

PR #450, opened 2026-05-30
refactor(proxmox_firewall): rename modules to clarify usage

PendaGTP:
##### SUMMARY

Rename modules which manage proxmox firewall options to more explicit name to prevent confusing with other incoming modules defined here #300

Noticed that a bit too late, sorry, I'm working on `proxmox_firewall_options` and noticed that.

`proxmox_firewall_options` -> options on VM/container-level
`proxmox_node_firewall_options` -> options at node-level
`proxmox_cluster_firewall_options` -> options at cluster-level
`proxmox_firewall` -> will be deprecated when #300 will be completed

##### ISSUE TYPE

- Refactor Pull Request

##### COMPONENT NAME

`proxmox_node_firewall`
`proxmox_node_firewall_info`
`proxmox_cluster_firewall`

##### ADDITIONAL INFORMATION

I hope we can do these changes, I’m not sure if it’s considered as a breaking change.

This is mainly because we already use `proxmox_firewall` and so we can't use it for vm/container level.

If this is a breaking change, and it's okay to have `proxmox_firewall_options` (vm/ct) and a naming missmatch with cluster and node fw options, we can close this PR.


IamLunchbox:
While this would make the naming convention more canonical, this is definetely not a minor change :D 

The renaming would break exisiting playbooks which rely on the old name. I did not find a way to alias a module, but that would be the best way to go. Create a alias mapping and throw a deprecation warning when a module is accessed by its old name.

Thulium-Drake:
You can re-route modules in `meta/runtime.yml`, here's how the folks at `community.general` do it: https://github.com/ansible-collections/community.general/blob/main/meta/runtime.yml#L83

But the bigger question is, how do we release this (and when), deprecations like this often times warrant a new major version. And as we 'just' released 2.0.0, that'd mean another (from the user's perspective) big bump. I'm not too sure that we should do that.

What we _can_ do however, is park the actual removal of the old/current names to 3.0.0 and pick it up when we get there.

PendaGTP:
Thanks for yours feedback! 

I also think that having a new major right now will be consufing for users.

Maybe we could release a new version 2.1.0 with routing and setting removal_version to 3.0.0? Idk if it's easy to release a minor version of collection or if this requires some background works.

IamLunchbox:
The minor release is not too much work AFAIK. But I think it would be sensible to keep the deprecation warnings down until 3.0.0 is in sight. I think it would be annoying for users to be warned of a deprecation, which is in one year? 🤔

PendaGTP:
I think that if it's annoying for users, they just have to migrate to the new naming which is not a big deal but I don't want forcing my ideas so obviously everyone opinion matters.

Also note that we shouldn't mark `proxmox_firewall` as deprecated but only `proxmox_node_firewall`, `proxmox_node_firewall_info` and `proxmox_cluster_firewall`.

- proxmox_node_firewall -> proxmox_node_firewall_options
- proxmox_node_firewall_info -> proxmox_node_firewall_options_info
- proxmox_cluster_firewall -> proxmox_cluster_firewall_options

PendaGTP:
I have added modules routing and deprecation notices (for testing of the behavior and prepare this PR if we decide to handle it before 3.0.0).

But I have tested localy and I don't see any warning message when using old modules names. Maybe this configuration isn't correct. But I have that inspiration from `community.general` and this seems to be the way to do this. Any ideas?

EDIT: fixed with suggestion applied!

meta/runtime.yml
```diff
@@ -70,3 +70,21 @@ action_groups:
     - proxmox_vnet_info
     - proxmox_zone
     - proxmox_zone_info
+
+plugin_routing:
+  modules:
+    community.proxmox.proxmox_node_firewall:
```
Thulium-Drake:
```suggestion
    proxmox_node_firewall:
```

I think the paths should be 'local'

meta/runtime.yml
```diff
@@ -70,3 +70,21 @@ action_groups:
...
+      redirect: community.proxmox.proxmox_node_firewall_options
+      deprecation:
+        removal_version: 3.0.0
+        warning_text: Use community.proxmox.proxmox_node_firewall_options instead.
+    community.proxmox.proxmox_node_firewall_info:
```
Thulium-Drake:
```suggestion
    proxmox_node_firewall_info:
```

meta/runtime.yml
```diff
@@ -70,3 +70,21 @@ action_groups:
...
+      redirect: community.proxmox.proxmox_node_firewall_options_info
+      deprecation:
+        removal_version: 3.0.0
+        warning_text: Use community.proxmox.proxmox_node_firewall_options_info instead.
+    community.proxmox.proxmox_cluster_firewall:
```
Thulium-Drake:
```suggestion
    proxmox_cluster_firewall:
```

meta/runtime.yml
PendaGTP: Thanks! 

This works:

```txt
redirecting (type: modules) community.proxmox.proxmox_node_firewall_info to community.proxmox.proxmox_node_firewall_options_info
```

IamLunchbox (approved):

Thulium-Drake:
<3 thanks for this work! :grin:

PR #455, opened 2026-06-04
docs(contributing): add welcoming intro

PendaGTP:
##### SUMMARY

Add a welcoming intro on contributing guide. The testing and AI-contribution rules can feel "intimidating" for first time contributors. This will I hope encourages people to ask for help, open WIP PRs or create an issues when stuck.

##### ISSUE TYPE

- Docs Pull Request

PR #456, opened 2026-06-04
docs(contributing): add pull request template

PendaGTP:
##### SUMMARY

Add a more detailed pull request template to helps contributors and reviewers efficency.

If we have to keep the default one from `ansible-collections` for consistency/norms that's ok to me and we can close this PR.

Relates #455 

##### ISSUE TYPE

- Docs Pull Request

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -23,7 +23,7 @@ Leave the [ ] if they are not applicable, or if you have not completed the item.
 - [ ] I have run unit tests.
 - [ ] I have run integration.
 - [ ] I have considered backward compatibility.
-- [ ] I haved added a changelog fragment.
+- [ ] I haved added a changelog fragment (expect for new a module).
 
 <!---
 You can find more information about coding conventions and local testing in the [CONTRIBUTING.md](https://github.com/ansible-collections/community.proxmox/blob/main/CONTRIBUTING.md) file.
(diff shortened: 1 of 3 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,33 @@
...
+<!---
+Please mark the following items with an [x] *IF* they apply to your PR.
+Leave the [ ] if they are not applicable, or if you have not completed the item.
+--->
+- [ ] I have run `nox` and fixed any issues.
+- [ ] I have added / updated unit tests (**required** for new modules and bug fixes).
+- [ ] I have run unit tests.
+- [ ] I have run integration.
+- [ ] I have considered backward compatibility.
+- [ ] I haved added a changelog fragment.
```
IamLunchbox: Maybe highlight, that we not need it for new modules?

IamLunchbox (commented):
I've seen Interactive PR templates as well, what do you think about that?

Otherwise: Nice work! 🙂

PendaGTP:
> I've seen Interactive PR templates as well, what do you think about that?

I not a big fan of this format, this is less markdown friendly and flexible. I would prefer to stick to "old fashion" way which is also more common and people are more used to use it.

This is more a personnal preference, idk more generaly what people think of it.

Thulium-Drake:
Let's start with the non-interactive one and see how well it's being used :-)

PR #457, opened 2026-06-05
test(integration): add helper script to run tests on dockerized proxmox

PendaGTP:
##### SUMMARY

Add an helper script to run integration tests on dockerized Proxmox.

This use this project [rtedpro-cpu/dockermox](https://github.com/rtedpro-cpu/dockermox) and these [docker images](https://hub.docker.com/r/rtedpro/proxmox)

Usage:

```txt
Usage: run.sh [--version VERSION] [--target TARGET] [--reuse] [--rm] [--prune]

Run integration tests against a Proxmox VE container.

Options:
  --version VERSION   Proxmox version (major like 9, exact like 9.2.3, or latest)
  --target TARGET     ansible-test integration target (e.g. proxmox_pool)
  --reuse             Reuse existing container if present
  --rm                Remove container after tests
  --prune             Remove the Docker image (implies --rm)

Defaults:
  --version           latest supported Proxmox version (from matrix.yml)
  --target            all integration test targets
  --reuse             off (recreate container if one already exists)
  --rm                off (keep container after tests)
  --prune             off (keep Docker image)

Examples:
  ./tests/integration/run.sh
  ./tests/integration/run.sh --target proxmox_pool
  ./tests/integration/run.sh --version 9 --target proxmox_pool
  ./tests/integration/run.sh --reuse --target proxmox_pool
  ./tests/integration/run.sh --rm --prune
```

Relates #7

##### ISSUE TYPE

- Test Pull Request

##### ADDITIONAL INFORMATION

Before merging this PR it could be nice to decide which PVE versions we should try to support the best. And so adjust the `matrix.yml` file.

I will update https://github.com/ansible-collections/community.proxmox/pull/456 / vice versa this one to reflect changes once any merged.

Currently some of our tests are outdated are broken, but can you test it with: `run.sh --target proxmox_pool`

One improvement could be to support also `podman` let me know if interested!


PendaGTP:
I have done some little adjustement:

- container name: "pve-integration-\<version\>"
- document defaults values in usage
- show usage when unknow param provided

CI fail but not related this these changes but due to a timeout `HTTP error 504 Server Error: Gateway Time-out for url: https://github.com/ansible/ansible/archive/stable-2.21.tar.gz`

Thulium-Drake:
This looks great! I checked the `dockur/proxmox` page, and it seems to also tag it's containers nicely with `latest` and whatever major and minor versions are used.

So maybe we should do the following, I'm always assuming folks run the latest supported version of hypervisors (as they are important to keep secure ;-) ). So let's start by always running the integration pipeline against `9`, which will be the latest supported release for that major. They don't provide `8`, but chances are they will start publishing `10` when we get there.

I don't see a lot of merit of testing all minor (let alone patch) releases, as it's more work for us to maintain the integration tests, while the solution to some of the bugs is probably a `apt update; apt dist-upgrade -y` away.

PendaGTP:
Container images are pushed to dockerhub and GH. On GH at this moment, there is only latest minor version available and on DockerHub there are all versions defined in matrix.yaml. Available versions has to be bump manualy on new release.

We can switch to use GH as registry and so use latest or 9 tag (which target latest minor) but having the possibility to test something on v8 can still be helpfull.

Same as you, I don't really see the need to test all minor version but sometime needed (e.g for https://github.com/ansible-collections/community.proxmox/pull/438).

We could document in CONTRIBUTING.md that integration test has to be run at least on latest version and other versions are for debugging purpose or specific needs. (Default behavior of the script run on latest)

Thulium-Drake:
Yes, I tend to agree with you, _however_, it seems that Proxmox will be dropping support for `8` in 10 weeks... https://endoflife.date/proxmox-ve / https://pve.proxmox.com/wiki/FAQ

So maybe in this case, we should just skip testing on 8 for now (as we've been running for a while without integration tests :-) )

PendaGTP:
Oh nice, I wasn't aware of that.

Ah second "Oh".., sorry I just realized that I set the wrong GH repo in the PR description:

There are two main project which dockerized proxmox:

- https://github.com/rtedpro-cpu/dockermox (https://hub.docker.com/r/rtedpro/proxmox)
- https://github.com/dockur/proxmox (https://hub.docker.com/r/dockurr/proxmox/tags)

`rtedpro-cpu/dockermox` only support LXC (this is the one I'm using on the helper)
`dockur/proxmox` support both but requires KVM acceleration

Currently I haven't done lot of tests to compare them. I'm using `rtedpro-cpu/dockermox` since my laptop doesn't support KVM, but I already PoC `dockur/proxmox` on an other host. Right now idk which path we should go.

PendaGTP:
I have migrated the script to use **dockur/proxmox** which simplify the version management (we don't need anymore a matrix file), support CT/VM and seems "more" maintained/used.

Let me know what do you think I can still revert that.

IamLunchbox:
The work of the dockur organization seems to be widely accepted. 

But since we offer to automate integration tests for other users, we must be aware, that this must be done with great care. Docker is no strict security boundary and if an image we trust gets compromised, this might result in a risk for users / testers of this project.

Therefore, maybe we could tag the image version hash in the script to make sure users don't download a malicious container? Additionally, I am a bit opposed to run a third party container in privileged mode, it could compromise the system more easily. But I just asked in the upstream repo if we can limit down the permissions to have a less risky integration test script.

Additionally, your script runs `set -e`, which will stop the script if a nonzero exit code is encountered. [This line](https://github.com/ansible-collections/community.proxmox/pull/457/changes#diff-f3170d300281b1f168e042fd806717eb7cff323632f631a6eac7e74272402cabR218) might actually stop the script before cleaning up though. :)

PendaGTP:
Thanks for you feedback. 

So I guess I should keep this script for my personnal usage. I can't ensure that the docker image is safe to use and doesn't impact in any way the host. Pinning to image version hash is possible, user can pass the hash on `--version`. Honestly if this doesn't looks good to integrate that because of concerns about responsability/security/etc.. It's ok to me. Let me know and I close this PR.

Thulium-Drake:
I think it's fine to merge it and make it generally available to all developers of the collection. Pinning the hash is a good plan, so unless Docker breaks it's registry, we should be able to trust that image without it changing.

Also, we could put this in the contributor guidelines as well, that we (the authors of this collection) _strongly_ recommend that you do not use your own personal workstation to run these containers on. And instead run them inside a VM or isolated hardware (most of the functionality doesn't need KVM acceleration anyway if you only want to play with the API).

And it might be useful to figure out a means of having multiple dockers so (some) of the cluster functionality is also available. But that could also be parked for a later iteration ;-)

PendaGTP:
Thanks for you reply!

I have added support of pinning image version to a digest and also add a note on the contributing doc to warn and encourage pin version.

Thulium-Drake (approved):
Thanks a lot, I like it! I think this will make developing modules a lot easier ;-)

IamLunchbox:
Since we seem to be settled to accept this script I'd have two ideas:
- podman support
- support to precede the command with sudo

What do you think?

PendaGTP:
> support to precede the command with sudo

You mean running docker command with sudo if needed (by auto detect or arg for example)? Personnaly, I don't think this is really valuable, adding the current user to docker group is the standard docker install method.

IamLunchbox:
Is it? I always run it with sudo 😅

I mean, forcing users to run the whole script with sudo is less elegant then elevating only the docker cmd if necessary

Thulium-Drake:
Also for `podman` support, all commands should be interchangeable, so testing for podman and docker binaries and use whichever can be found first should be fine. I don't expect any changes to the script to make it work.

Or you can do `alias docker=podman` in your shell ;-)

PendaGTP:
Done :heavy_check_mark:

Thulium-Drake:
@IamLunchbox LGTM, Do you agree? :grin:

PR #469, opened 2026-07-05
Add `local` connection support.

### LINKED ISSUE #38: Support proxmoxer local mode

### LINKED ISSUE #348: Add `local` connection support.

PendaGTP:
### What does this PR do?

Continues work from #349 by @reitermarkus, who implemented `local` connection. This PR was blocked on lint/merge conflict issue.

My changes: format and resolve merge conflict

### Issue Type

- Feature Pull Request

### Component Name

`proxmox`

### Contributor's Note

- [x] I have run unit tests.
- [ ] I have run integration.

Fixes #348 
Fixes #38

### DIFF
--- a/changelogs/fragments/349-local-backend.yml
+++ b/changelogs/fragments/349-local-backend.yml
@@ -1,0 +1,5 @@
+---
+minor_changes:
+  - proxmox - make ``api_host`` optional to support ``local`` backend (https://github.com/ansible-collections/community.proxmox/pull/349).
+  - proxmox - make ``api_user`` optional and default to ``root@pam`` (https://github.com/ansible-collections/community.proxmox/pull/349).
+  - proxmox - use ``local`` backend if no ``api_host`` is specified (https://github.com/ansible-collections/community.proxmox/pull/349).
--- a/plugins/doc_fragments/proxmox.py
+++ b/plugins/doc_fragments/proxmox.py
@@ -12,19 +12,17 @@ options:
       - Specify the target host of the Proxmox VE cluster.
       - Uses the E(PROXMOX_HOST) environment variable if not specified.
     type: str
-    required: true
   api_port:
     description:
       - Specify the target port of the Proxmox VE cluster.
       - Uses the E(PROXMOX_PORT) environment variable if not specified.
     type: int
-    required: false
   api_user:
     description:
       - Specify the user to authenticate with.
       - Uses the E(PROXMOX_USER) environment variable if not specified.
     type: str
-    required: true
+    default: root@pam
   api_password:
     description:
       - Specify the password to authenticate with.
--- a/plugins/module_utils/proxmox.py
+++ b/plugins/module_utils/proxmox.py
@@ -32,9 +32,9 @@ def proxmox_auth_argument_spec():
         dict[str, dict]: Parameter names mapped to their configuration dictionaries.
     """
     return dict(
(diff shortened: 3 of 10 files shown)


Thulium-Drake (approved):
@PendaGTP Thanks for picking this up :heart:

IamLunchbox:
I will check it out tomorrow and probably merge then :)

PR #491, opened 2026-07-31
fix(inventory): pass verify= on requests when validate_certs is set

aioue:
## Summary

Homelab and CI setups often set `validate_certs: false` in the Proxmox inventory plugin when connecting to PVE's self-signed HTTPS API. That option is supposed to skip certificate verification, but in many environments it silently does nothing.

**What users see:** `ansible-inventory` fails with `SSL: CERTIFICATE_VERIFY_FAILED` even though `validate_certs: false` is set in the inventory file.

**Why:** The plugin only set `session.verify` on the shared `requests` session. When `REQUESTS_CA_BUNDLE` is set in the environment (common in devcontainers, CI images, and hardened Python installs), recent `requests`/`urllib3` still verify TLS and ignore `session.verify=False`. You have to pass `verify=` on each request.

**Fix:** Read `validate_certs` once and pass `verify=` explicitly on the inventory plugin's `GET` and `POST` calls (password auth ticket request included). No new options; behaviour matches what the docs already promise.

Same class of fix as ansible-collections/community.general#9178.

**Note:** This does not address hostname mismatch (e.g. inventory `url` uses an IP but the node cert is issued for `pve.localdomain`). That still requires the correct hostname or a trusted cert for the URL you use. Only the "skip verification" path is fixed here.

## Test plan

- [x] Reproduced: `validate_certs: false` fails with `REQUESTS_CA_BUNDLE=/etc/ssl/certs/ca-certificates.crt` before patch
- [x] After patch: `ansible-inventory` succeeds with `validate_certs: false` and `REQUESTS_CA_BUNDLE` set
- [x] `validate_certs: true` unchanged where verification already worked

### DIFF
--- a/changelogs/fragments/inventory-validate-certs-requests-ca-bundle.yml
+++ b/changelogs/fragments/inventory-validate-certs-requests-ca-bundle.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - "proxmox inventory - pass ``verify`` explicitly on HTTP requests so ``validate_certs: false`` is honored when ``REQUESTS_CA_BUNDLE`` is set (regression when only ``session.verify`` was used) (https://github.com/ansible-collections/community.proxmox/pull/491)."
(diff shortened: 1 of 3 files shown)


IamLunchbox:
Thank you for the fix. :rocket: 

That comes from intentory not using proxmoxer. 

Anyways, I think your changelog fragment is off, it gets interpreted as a dict. :thinking:

aioue:
Wasn't even aware of proxmoxer until just now!

IamLunchbox (approved):

Thulium-Drake:
Thanks for this fix! :rocket:

PR #492, opened 2026-08-01
fix(proxmox_pct_remote): restore missing _raise_paramiko_connect_exception helper

### LINKED ISSUE #477: proxmox_pct_remote connection uses undefined `_raise_paramiko_connect_exception()` method
##### SUMMARY
In case of exceptions, the connection plugin calls `self._raise_paramiko_connect_exception()`, which does not exist.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
proxmox_pct_remote connection plugin

viniciussaporto:
## SUMMARY

`proxmox_pct_remote` calls `self._raise_paramiko_connect_exception(e, port)` in
its generic connect-error handler, but that method does not exist anywhere in
the collection. `git log -S` pins its disappearance to 0ffe419 (PR #372), a
ruff style refactor that extracted `_paramiko_connect_ssh()` and replaced the
inline error-translation block with a call to a helper it never added.

Any connect error other than a bad host key or an authentication failure,
connection refused, timeouts, `NoValidConnectionsError`, DNS failures,
therefore crashes with `AttributeError`, hiding the real reason.

It also misclassifies the host: because `AttributeError` propagates as an
ordinary exception rather than `AnsibleConnectionFailure`, ansible-core records
the host as failed instead of unreachable, which changes behaviour for
`ignore_unreachable`, `max_fail_percentage`, `serial`, and retry files.

This restores the helper with the pre-refactor behaviour, using `get_option()`
rather than the original `get_options()`, the latter returns the whole options
dict and would have rendered the password into the error message.

Reproduced and verified end-to-end against a real Proxmox VE 9.1.0 node with a
Debian 13 LXC, before/after logs are attached to the issue.

Fixes #477

## ISSUE TYPE

- Bugfix Pull Request

## COMPONENT NAME

proxmox_pct_remote

## ADDITIONAL INFORMATION

I've discussed on #477 before opening this PR. Three unit tests cover the restored
behaviour, verified red (`3 failed, 31 passed` against upstream's copy of the
plugin) then green (`34 passed`). `ansible-test sanity` and the repository
pre-commit hooks pass. Changelog fragment included.

One point raised on the issue and carried here:

I used `get_option()` where the pre-#372 code used `get_options()`. That is
a deliberate deviation from a literal restoration, since the original would
have leaked every connection option including the password into the error
text. A unit test asserts the password does not appear.

### DIFF
--- a/plugins/connection/proxmox_pct_remote.py
+++ b/plugins/connection/proxmox_pct_remote.py
@@ -595,19 +595,6 @@ class Connection(ConnectionBase):
             self._raise_paramiko_connect_exception(e, port)
 
     def _raise_paramiko_connect_exception(self, e: Exception, port: int) -> t.NoReturn:
-        """Translate an unexpected Paramiko connect error into an Ansible failure.
-
-        Restores the error translation that was inlined in ``_connect`` before
-        the refactor in #372 extracted ``_paramiko_connect_ssh`` and referenced
-        this helper without defining it (issue #477). Without it, any
-        unexpected connection error (e.g. connection refused, timeouts,
-        ``NoValidConnectionsError``) crashed with ``AttributeError`` instead of
-        reporting the real failure.
-
-        Uses ``get_option`` (singular); the pre-refactor code called
-        ``get_options('remote_addr')``, which would have rendered the entire
-        options dict - including the password - into the error message.
-        """
         msg = to_text(e)
         if "PID check failed" in msg:
             raise AnsibleError("paramiko version issue, please upgrade paramiko on the machine running ansible") from e
--- a/plugins/connection/proxmox_pct_remote.py
+++ b/plugins/connection/proxmox_pct_remote.py
@@ -594,6 +594,30 @@ class Connection(ConnectionBase):
         except Exception as e:
             self._raise_paramiko_connect_exception(e, port)
 
+    def _raise_paramiko_connect_exception(self, e: Exception, port: int) -> t.NoReturn:
+        """Translate an unexpected Paramiko connect error into an Ansible failure.
+
+        Restores the error translation that was inlined in ``_connect`` before
+        the refactor in #372 extracted ``_paramiko_connect_ssh`` and referenced
+        this helper without defining it (issue #477). Without it, any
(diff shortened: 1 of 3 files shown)


IamLunchbox (approved):

viniciussaporto:
Took some days, I was busy with work and life.

CI was red on PLR0917 for test_tempfile_creation_and_move. It's pre-existing on main, the signature already carried noqa: PLR0913, and newer ruff (0.16.1) additionally flagged PLR0917 for the same positional arguments. It only showed up here because change detection lints the entire file once it's touched.

Pushed a one line commit extending the existing noqa to cover both rules.

Worth noting the repo .pre-commit-config.yaml pins ruff v0.15.9, which doesn't flag this, so local pre-commit runs won't match CI here. Might be worth bumping the pin, your guy's choice!

Thanks and happy weekend!

plugins/connection/proxmox_pct_remote.py
```diff
@@ -594,6 +594,30 @@ def _paramiko_connect_ssh(
...
+        Restores the error translation that was inlined in ``_connect`` before
+        the refactor in #372 extracted ``_paramiko_connect_ssh`` and referenced
+        this helper without defining it (issue #477). Without it, any
+        unexpected connection error (e.g. connection refused, timeouts,
+        ``NoValidConnectionsError``) crashed with ``AttributeError`` instead of
+        reporting the real failure.
+
+        Uses ``get_option`` (singular); the pre-refactor code called
+        ``get_options('remote_addr')``, which would have rendered the entire
+        options dict - including the password - into the error message.
```
felixfontein: That's information that belongs into a PR description or commit message, but not into a code comment IMO.
```suggestion
```

felixfontein (commented):
Besides the comment, LGTM!

plugins/connection/proxmox_pct_remote.py
IamLunchbox: I tend to agree :+1:

viniciussaporto:
Hi, changed the comment to the commit message. Thanks for the tips, guys!

Thulium-Drake:
Thanks for this fix! :rocket:

PR #504, opened 2026-08-16
fix(proxmox_vm_info): skip VM network info when VM is stopped

### LINKED ISSUE #495: proxmox_vm_info should have a non-crashing mode for being able to retrieve maximum infos from all VMs
##### SUMMARY

Currently, `proxmox_vm_info` crashes when it tries to retrieve infos such as with `network: true` which requires VMs to be running or the guest agent being available. This is great when working with single VMs, but bad when working with many VMs where I'm more interested to detect their current state.

Then following errors might be thrown:
```
TASK [update-proxmox : Gather VM information from Proxmox] ********************************************************************************************************************************************************************
[ERROR]: Task failed: Module failed: Failed to retrieve QEMU VMs information: 500 Internal Server Error: VM 801 is not running
Origin: /home/zocker/projects/server/ansible/roles/update-proxmox/tasks/_gather-vm-facts.yml:3:3

1 ---
2
3 - name: Gather VM information from Proxmox
    ^ column 3

fatal: [pve -> localhost]: FAILED! => {"changed": false, "msg": "Failed to retrieve QEMU VMs information: 500 Internal Server Error: VM 801 is not running"}
```

```
TASK [Gather infos from single VM] ********************************************************************************************************************************************************************************************
[ERROR]: Task failed: Module failed: Failed to retrieve QEMU VMs information: 500 Internal Server Error: QEMU guest agent is not running
Origin: /home/zocker/projects/server/ansible/site.yml:15:7

13   become: false
14   tasks:
15     - name: Gather infos from single VM
         ^ column 7

fatal: [pve -> localhost]: FAILED! => {"changed": false, "msg": "Failed to retrieve QEMU VMs information: 500 Internal Server Error: QEMU guest agent is not running"}
```

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

proxmox_vm_info

##### ADDITIONAL INFORMATION

In my case, I want to use the info whether the guest agent is connected to derive whether a VM is fully booted or not (as kind of a smoke test). This seems currently not possible because there is currently no value which reflects this status. When with `network: true` the module would not crash on disabled VMs but just pass this info to me as `"network": null`, then I would work with this.

PendaGTP:
### What does this PR do?

Skip gather VM network information when the VM is stopped.

### Issue Type

- Bugfix Pull Request

### Component Name

`proxmox_vm_info`

### Contributor's Note

- [ ] I have added / updated unit tests (**required** for new modules and bug fixes).
- [x] I have run unit tests.
- [x] I have run integration.

Fixes #495

### DIFF
--- a/changelogs/fragments/504-vm-network-info-when-stopped.yml
+++ b/changelogs/fragments/504-vm-network-info-when-stopped.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - proxmox_vm_info - skip VM network info when VM is stopped (https://github.com/ansible-collections/community.proxmox/pull/504).
--- a/plugins/modules/proxmox_vm_info.py
+++ b/plugins/modules/proxmox_vm_info.py
@@ -200,13 +200,13 @@ class ProxmoxVmInfoAnsible(ProxmoxAnsible):
                         config_type = 0 if config == "pending" else 1
                         # GET /nodes/{node}/qemu/{vmid}/config current=[0/1]
                         desired_vm["config"] = call_vm_getter(this_vm_id).config().get(current=config_type)
-                    if network:
-                        if resource_type == "qemu":
-                            desired_vm["network"] = (
-                                call_vm_getter(this_vm_id).agent("network-get-interfaces").get()["result"]
-                            )
-                        elif resource_type == "lxc":
-                            desired_vm["network"] = call_vm_getter(this_vm_id).interfaces.get()
+                        if network and desired_vm["status"] == "running":
+                            if resource_type == "qemu":
+                                desired_vm["network"] = (
+                                    call_vm_getter(this_vm_id).agent("network-get-interfaces").get()["result"]
+                                )
+                            elif resource_type == "lxc":
+                                desired_vm["network"] = call_vm_getter(this_vm_id).interfaces.get()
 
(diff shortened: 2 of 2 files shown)


PendaGTP:
For lint issues, see #505

Thulium-Drake (approved):
<3

Zocker1999NET:
thx for fixing on this so quickly :heart: 

unfortunately, this does not fully fix the problem I tried to described in my issue (#495), because while the module now no longer crashes on powered off VMs, it still fails with an error when the VM is running but the QEMU agent isn't, and thats the use case I tried to describe at the end. (I have tested this with the current main branch.)

I am grateful through as now the module will only crash my playbook when I intend my playbook to crash anyway (through I planned it with a better message).

If you say how you would prefer it (additional option or just silence error), I can extend this fix for my own use case & push it here, if you want.

PendaGTP:
Thanks for your feedback, appreciate!

You can open an new issue to explain that there is an error when guest agent are not running. And you’re more than welcome to open a pull request to implement a fix, otherwise, I could also take care of it, just let me know.

PendaGTP:
@Zocker1999NET I open an PR for it https://github.com/ansible-collections/community.proxmox/pull/508 (having free time this morning)

PR #506, opened 2026-08-17
Add support for mounting passthrough devices on LXCs

### LINKED ISSUE #144: Add support for device mount points
##### SUMMARY

There is only way to mount block devices of the host directly into the LXC - use `mounts` but it's mutually exclusive with `mount_volumes`.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

proxmox

##### ADDITIONAL INFORMATION

Add `mount_devices` to create device mount points.

```yaml
    mount_devices:
      - id: dev0
        path: /dev/dri/renderD128
        deny-write: false
        gid: 1234
        mode: 755
        uid: 1234
```

### LINKED ISSUE #458: Device pass through to container
##### SUMMARY
Add parameter to pass through devices to the container

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
Module: proxmox

##### ADDITIONAL INFORMATION
I'd really appreciate the ability to pass through device(s) to the container.

From the [Proxmox API documentation](https://pve.proxmox.com/pve-docs/api-viewer/index.html#/nodes/{node}/lxc/{vmid}/config)
Name: dev[n]
Type: string
Format: [[path=]<Path>] [,deny-write=<1|0>] [,gid=<integer>] [,mode=<Octal access mode>] [,uid=<integer>]

emanguy:
### What does this PR do?
This PR resolves #144 and #458 to add a new "mount_devices" argument for `community.proxmox.proxmox` to allow adding passthrough devices into LXCs.

The only caveat (which I've documented) is that the proxmox API only allows changes to passthrough devices via username/password authentication with root@pam.

I verified this works by running this playbook against my personal proxmox instance and verifying that the VM boots and has the mounted device in its /dev directory:

```yaml
- hosts: olympus
  module_defaults:
    # I've aliased my local copy of the collection as "community.proxmox_test"
    group/community.proxmox_test.proxmox:
      api_host: "{{ proxmox_api_host }}"
      api_user: "{{ proxmox_api_user }}"
      api_password: "{{ proxmox_api_pass }}"
      validate_certs: false
  tasks:
    - name: Start VM with forwarded devices
      community.proxmox_test.proxmox:
        vmid: 200
        node: olympus
        password: notarealpassword
        hostname: ansibletest
        ostemplate: vm-templates:vztmpl/ubuntu-26.04-standard_26.04-1_amd64.tar.zst
        ostype: ubuntu
        memory: 1024
        swap: 512
        disk_volume:
          size: 16
          storage: vm-disks
        mount_devices:
          - id: dev0
            path: /dev/nvidia0
        nameserver: 10.0.0.1
        netif:
          net0: "name=eth0,bridge=vmbr0,firewall=1,ip=dhcp,type=veth"
          net1: "name=internal0,bridge=vmbr1,firewall=1,ip=dhcp,type=veth"
        features:
          - nesting=1
    - name: Configure VM firewall
      community.proxmox_test.proxmox_firewall:
        level: vm
        vmid: 200
        rules:
          - type: in
            action: ACCEPT
            source: "10.0.0.0/16"
            pos: 10
            dport: 80
            proto: tcp
```

### Issue Type

- Feature Pull Request

### Component Name

community.proxmox.proxmox - passthrough device mounting support

### Contributor's Note

- [x] I have run unit tests.
- [ ] I have run integration. (I'll trust CI on this one)

Fixes #144
Fixes #458


IamLunchbox:
You can't trust our CI regarding integration. CI does not cover that. :(

Thank you for the PR, though! We'll look into that.

emanguy:
Gotcha - I'll try running it on my machine in a bit and let you know the results! Is there a way to configure a virtualenv for the control node to use? I'm running Ubuntu 26 which uses apt to manage system python dependencies - the official apt repos only have proxmoxer 2.2.

emanguy:
Running into a problem with running CI locally - I'm running `./tests/integration/run.sh` and the tests start, but it chokes at this line because the variables "user" and "domain" aren't being set:

https://github.com/ansible-collections/community.proxmox/blob/d3497a832be45bcaa8095edc55e67f363707ab5e/tests/integration/targets/proxmox/tasks/main.yml#L55-L56

I'm following the instructions in CONTRIBUTING.md, is there something I'm missing?

IamLunchbox:
Since you continued to commit: Did you fix the bug? I think in theory you could set domain through your hosts file 🤔 

But yeah, I check the file out :)

emanguy:
No, it seems to be a variable that gets referenced in the test playbook for the integration test but I don't see where it would be getting set, so ansible just fails when the integration tests start

emanguy:
Specifically, this is the error I get:

image

IamLunchbox:
Sorry, i will need to check it out but did not have the time yet. @PendaGTP, do you know maybe if this is a bug on our side?

PendaGTP:
@IamLunchbox

This is not a bug, this is "expected", once I have worked on integration tests, I haven't updated any tests and some were already failing. I just added the helper.

So, I think we can ignore it and handle it separetely.

I will open an issue which reference which tests works and doesn't so we can track this and work on this soon.

IamLunchbox:
Yeah, good point. 

@emanguy Would you take care of the CI error? I think we could merge then.

plugins/modules/proxmox.py
```diff
@@ -1633,6 +1687,77 @@ def check_supported_features(self):
...
+
+        Returns:
+            Dict[str, str]: Formatted dictionary of device ID to comma-separated path and arguments.
+        """
+        if devices is None:
+            return {}
+        adjusted_devices = {}
+        device_id_regex = re.compile(r"^dev\d+$")
+        for device_entry in devices:
+            if device_id_regex.match(device_entry["id"]) is None:
```
IamLunchbox, line 1706: Is it guaranteed, that the key id exists?

Otherwise, I'd prefer using device_entry.get("id").

plugins/modules/proxmox.py
```diff
@@ -1633,6 +1687,77 @@ def check_supported_features(self):
...
+                self.module.fail_json(msg=f"Device {device_entry['id']} - Invalid device ID")
+
+            if device_entry["id"] in adjusted_devices:
```
IamLunchbox, line 1709: See above^

IamLunchbox (approved):

plugins/modules/proxmox.py
emanguy: Both ID and path are defined with "required=True" as seen here: https://github.com/emanguy/community.proxmox/blob/bddbc5f60b3052ecc6df689d6532a9a4b160304d/plugins/modules/proxmox.py#L684-L685

Is that sufficient to assume it always exists or are there cases where the "required" check may fail?

emanguy:
Sure, working on the CI issues now, will notify you once they're addressed

emanguy:
Just applied formatting fixes from Ruff, both that and CodeQA pass locally:
image

As does the sanity tests:
image

CI should be good to go now! :crossed_fingers:

Thulium-Drake:
CI seems happy ;-)

changelogs/fragments/144-add-device-passthrough.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+  - proxmox - add ``mount_devices`` option to allow mounting passthrough devices for LXC containers. Note that Proxmox currently requires root@pam password login to manage passthrough devices (https://github.com/ansible-collections/community.general/pull/506).
```
emanguy: Don't merge yet! This PR link is wrong

emanguy: Crisis averted, the PR link is correct now.

PR #512, opened 2026-09-01
fix(proxmox_node_firewall_options): apply state=disabled without host.fw

### LINKED ISSUE #511: proxmox_node_firewall_options: state=disabled does nothing on a node without host.fw
##### SUMMARY

`state: disabled` silently does nothing on a node that has no `/etc/pve/nodes/<node>/host.fw` file.
The task reports `changed: false` and `msg: "Node firewall options already match desired state"`.

But Proxmox VE treats a missing `host.fw` as enabled by default:

1. a missing file parses to `{}`:  https://github.com/proxmox/pve-firewall/blob/8a1b22c9c40da0112b40c4723edeba1181fb239e/src/PVE/Firewall.pm#L3456-L3462
2. that turns into an empty options hash: https://github.com/proxmox/pve-firewall/blob/8a1b22c9c40da0112b40c4723edeba1181fb239e/src/PVE/Firewall.pm#L4348
3. and the host firewall is switched on unless `enable` is present *and* it is `0`: https://github.com/proxmox/pve-firewall/blob/8a1b22c9c40da0112b40c4723edeba1181fb239e/src/PVE/Firewall.pm#L4378

The Rust firewall config library (behind the nftables backend) has the same default:

1. `HOST_ENABLED_DEFAULT: bool = true`: https://github.com/proxmox/proxmox-ve-rs/blob/61125e1fbf8d0f1a6385153bfd6eae36254b9b89/proxmox-ve-config/src/firewall/host.rs#L17-L18
2. `is_enabled()` falls back to the default: https://github.com/proxmox/proxmox-ve-rs/blob/61125e1fbf8d0f1a6385153bfd6eae36254b9b89/proxmox-ve-config/src/firewall/host.rs#L160-L162

But `proxmox_node_firewall_options` defines the default as disabled:
https://github.com/ansible-collections/community.proxmox/blob/5d5079acf216f039f29a192910c3afd1d7bb7c59/plugins/module_utils/proxmox_node_firewall_options.py#L11-L16

A node without `host.fw` returns no options at all, just the digest of the empty file:
`{"digest": "da39a3ee5e6b4b0d3255bfef95601890afd80709"}`.
This is because `GET /nodes/{node}/firewall/options` never fills in schema defaults:
https://github.com/proxmox/pve-firewall/blob/8a1b22c9c40da0112b40c4723edeba1181fb239e/src/PVE/API2/Firewall/Host.pm#L89

The module fills that missing API field with its own default.
`_is_update_required()` then compares `enable: 0` against `enable: 0`,
finds no difference, and sends no `PUT`:
https://github.com/ansible-collections/community.proxmox/blob/5d5079acf216f039f29a192910c3afd1d7bb7c59/plugins/modules/proxmox_node_firewall_options.py#L440-L441

**Impact**
a node that joins a cluster with the firewall already enabled has no `host.fw`.
PVE filters its host traffic with the cluster-wide default `policy_in: DROP`.
Running this module with `state: disabled` reports success but changes nothing.

##### ISSUE TYPE

- Bug Report

##### COMPONENT NAME

proxmox_node_firewall_options

Also `proxmox_node_firewall_options_info`, which misreports the same state.

On the 2.0.0 release these are `proxmox_node_firewall` / `proxmox_node_firewall_info`.
Both 2.0.0 and latest `main` (5d5079a) are affected.

##### ANSIBLE VERSION

```text
ansible [core 2.19.11]
  config file = /home/ci/workspace/ansible/ansible.cfg
  configured module search path = ['/home/ci/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /opt/virtualenvs/ansible-12.3/lib/python3.13/site-packages/ansible
  ansible collection location = /home/ci/.ansible/collections:/usr/share/ansible/collections
  executable location = /opt/virtualenvs/ansible-12.3/bin/ansible
  python version = 3.13.14 (main, Jun 11 2026, 00:00:00) [GCC 15.2.1 20260123 (Red Hat 15.2.1-7)]
  jinja version = 3.1.6
  pyyaml version = 6.0.3 (with libyaml v0.2.5)
```

##### COLLECTION VERSION

```text
# /home/ci/.ansible/collections/ansible_collections
Collection        Version
----------------- -------
community.proxmox 2.0.0
```

##### CONFIGURATION

```text
ANSIBLE_PIPELINING = True
DEFAULT_BECOME = True
DEFAULT_FORKS = 30
DEFAULT_STRATEGY = mitogen_linear
HOST_KEY_CHECKING = False
INTERPRETER_PYTHON = auto
RETRY_FILES_ENABLED = False
```

##### OS / ENVIRONMENT

* Proxmox VE 9.2 (pve-manager/9.2.10)
* pve-firewall 6.0.5
* proxmox-firewall 1.2.3
* proxmoxer 2.3.0
* requests 2.34.2.
* Controller: Fedora 43 container, Python 3.13.

##### STEPS TO REPRODUCE

Add a node to the proxmox cluster that has never had its firewall options set

```yaml
- name: Disable the host firewall on a node that has no host.fw yet
  community.proxmox.proxmox_node_firewall_options:
    api_host: "{{ pve_api_host }}"
    api_user: root@pam
    api_password: "{{ pve_api_password }}"
    node_name: pve-001
    state: disabled
```

Or simply run the following test, e.g. `tests/unit/plugins/modules/test_proxmox_node_firewall_options.py`:

```python
    def test_disable_firewall_on_node_without_host_fw(self):
        self.mock_api_fw_options.get.side_effect = [
            {"digest": "da39a3ee5e6b4b0d3255bfef95601890afd80709"},
            {"enable": 0, "digest": "da39a3ee5e6b4b0d3255bfef95601890afd80709"},
        ]

        result = self._run_module(build_module_args(state="disabled"))

        assert result["changed"] is True
        assert self.mock_api_fw_options.put.called
        assert self.mock_api_fw_options.put.call_args[1]["enable"] == 0
```

##### EXPECTED RESULTS

The module writes `enable: 0` (+ creating `host.fw`) and reports `changed: true`.
`_info` reports `enabled: true` for a node with no `enable` option, matching what PVE enforces.

##### ACTUAL RESULTS

No `PUT` is issued to PVE and the file is never created.

Module output when the API returns what a real node without a `host.fw` returns (only the
digest, as shown above):

```json
{
  "changed": false,
  "enabled": false,
  "msg": "Node firewall options already match desired state",
  "node_name": "pve-001",
  "ndp": true,
  "nftables": false,
  "nosmurfs": true,
  "nf_conntrack_max": 262144,
  "nf_conntrack_tcp_timeout_established": 432000,
  "nf_conntrack_tcp_timeout_syn_recv": 60,
  ...
}
```

grische:
### What does this PR do?

Proxmox VE enables the host firewall unless `host.fw` states `enable: 0`, and the
API returns that file without filling in defaults. The module read a missing
`enable` as *disabled*. On a node that has no `host.fw` yet, `state: disabled`
therefore did nothing, and the node kept filtering host traffic.

- `state: enabled` on such a node now reports no change instead of writing the
  file. Proxmox VE already treats it as enabled, so there is nothing to do.
- `proxmox_node_firewall_options_info` reports `enabled: true` for such a node
  instead of `false`.

Details in #511.

### Issue Type
- Bugfix Pull Request

### Component Name
proxmox_node_firewall_options, proxmox_node_firewall_options_info

### Contributor's Note
- [ ] I have run `nox` and fixed any issues.
- [x] I have run unit tests.
- [ ] I have run integration.
- [x] I have considered backward compatibility - it "does" change the playbooks behaviour for hosts without a host.fw

Fixes #511

### DIFF
--- a/changelogs/fragments/511-node-firewall-options-enable-default.yml
+++ b/changelogs/fragments/511-node-firewall-options-enable-default.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - proxmox_node_firewall_options - apply ``state=disabled`` on a node whose ``host.fw`` does not set ``enable`` (https://github.com/ansible-collections/community.proxmox/issues/511).
+  - proxmox_node_firewall_options_info - report ``enabled`` as ``true`` for a node whose ``host.fw`` does not set ``enable`` (https://github.com/ansible-collections/community.proxmox/issues/511).
--- a/plugins/module_utils/proxmox_node_firewall_options.py
+++ b/plugins/module_utils/proxmox_node_firewall_options.py
@@ -10,7 +10,8 @@ from ansible_collections.community.proxmox.plugins.module_utils.proxmox import (
 SCHEMA = {
     "enabled": {
         "api": "enable",
-        "default": False,
+        # Proxmox VE enables the host firewall unless host.fw states 'enable: 0'
+        "default": True,
         "to_api": ansible_to_proxmox_bool,
         "from_api": proxmox_to_ansible_bool,
     },
--- a/tests/unit/plugins/modules/test_proxmox_node_firewall_options.py
+++ b/tests/unit/plugins/modules/test_proxmox_node_firewall_options.py
@@ -150,6 +150,29 @@ class TestProxmoxNodeFirewallOptionsModule(ModuleTestCase):
         assert payload["enable"] == 1
         assert payload["nftables"] == 1
 
+    def test_disable_firewall_on_node_without_host_fw(self):
+        self.mock_api_fw_options.get.side_effect = [{}, {"enable": 0}]
+
+        result = self._run_module(build_module_args(state="disabled"))
+
+        assert result["changed"] is True
+        assert result["msg"] == "Node firewall options updated"
+        assert result["enabled"] is False
+
+        assert self.mock_api_fw_options.put.called
+        payload = self.mock_api_fw_options.put.call_args[1]
+        assert payload["enable"] == 0
+
+    def test_idempotent_when_host_fw_only_disables_the_firewall(self):
+        self.mock_api_fw_options.get.return_value = {"enable": 0}
+
+        result = self._run_module(build_module_args(state="disabled"))
+
+        assert result["changed"] is False
+        assert result["msg"] == "Node firewall options already match desired state"
+        assert result["enabled"] is False
+        assert not self.mock_api_fw_options.put.called
+
     def test_disable_firewall_updates_enable_flag(self):
         self.mock_api_fw_options.get.side_effect = [
             SAMPLE_API_NODE_FIREWALL_OPTIONS,


Thulium-Drake (approved):

IamLunchbox:
I guess this is a breaking change, though? When the file exists the default changes. 

Therefore i'd rather suggest to use check if the dict is empty and PUT in that case. 

Additionally we could change the default...after deprecating the current behaviour :)

Thulium-Drake:
I'm not sure, from what I gathered from the description I don't think so, as it now assumes the firewall is enabled when it's undefined. 

@grische What happens after this change when you run it against a note that has the firewall disabled and you want to enable it? Does it report a `change`, or `ok` (no change)?

I'd say we'd still need to have that feedback from the module

grische:
I am not sure I fully understand the question, but if a `host.fw` has `enable: 0` the schema's default is not being used, so it's unchanged to the previous behaviour and triggers a "change".
If a `host.fw` has no enable key (or is missing entirely) the host reports "ok" as PVE considers this firewall to be enabled (assuming all other options match as well).

IamLunchbox:
Ok, lets do it then :)

PR #493, opened 2026-08-02
Omit "pos" for API that does not support it, closes #485

agross:
### What does this PR do?

`pos` is rejected by the Proxmox API when updating rules.

### Issue Type

- Bugfix Pull Request

### Component Name
community.proxmox.proxmox_firewall

### Contributor's Note

- [ ] I have run `nox` and fixed any issues.
- [ ] I have added / updated unit tests (**required** for new modules and bug fixes).
- [ ] I have run unit tests.
- [ ] I have run integration.
- [ ] I have considered backward compatibility.
- [ ] I haved added a changelog fragment (expect for new a module).

Fixes #485

PR #525, opened 2026-09-21
fix(proxmox_firewall): update a rule only once when several params differ

grische:
### What does this PR do?
`compare_list_of_dicts` appended an item to the update list once per differing parameter. `proxmox_firewall` then updated the same rule dict again, and since the first pass had cleared its `pos`, the second pass addressed the rule endpoint `None` and failed.

### Issue Type
- Bugfix Pull Request

### Component Name
proxmox_firewall, plugins/module_utils/proxmox.py

### Contributor's Note
- [x] I have run unit tests.
- [ ] I have run integration.

Relates #493


grische:
CI failure fixed in #526

